PR #164, opened 2014-09-24
Remove support for HTTP

ghost:
After reddit enabled sitewide SSL for every user there is no reason to continue to support HTTP.
In this way it's no longer possible to ignore HTTPS enabled by the user in the reddit preferences and RedReader will continue working for all users even if reddit becomes stricter with ignoring the SSL rules.


ajs124:
It seems like reddit doesn't [support http anymore](https://www.reddit.com/r/redditdev/comments/39zje0/reddit_will_soon_only_be_available_over_https/).

PR #202, opened 2015-01-26
Fixes #201: Search on a subreddit twice doesn't work correctly

### LINKED ISSUE #201: Search on a subreddit twice doesn't work correctly
Hello, the scenario is like this:
1) you are in a subreddit, you do a search
2) the search is correctly done in the current subreddit
3) now if in the result page you try to do another search, the new search loses the subreddit limitations, it's done on all reddit. I would expect that it is kept in order to be able to refine the previous search
Note that also if you try yo submit a post from the search result page, you don't have the current subreddit preselected
Thanks

chaitan94:
As said, second search on a subreddit is done on that subreddit itself, instead of a general search.


QuantumBadger:
Excellent, thanks!

Issue #99, opened 2013-08-06
Allow sorting search results


aguslr:
It'd be great if the search results could be sorted by relevance, new, hot, top, etc. This is possible in the website and I miss it in the app.

By the way, great app. Love the work!

mvy:
Hi! 

I thought I'd add a bump to this issue. The search without the ability to sort is barely useless. 
Every time I search for something in my subreddits, I want to look a comments that are new, and rarely the most popuplar/upvoted. 

M'vy

QuantumBadger:
Thanks to chaitan94 this has just been added and will be in the next version :)

https://github.com/QuantumBadger/RedReader/pull/203

PR #203, opened 2015-01-26
Implemented #99: Allow sorting search results

chaitan94:
Added the search specific sortings: relevance, new, hot, top, comments


QuantumBadger:
Brilliant stuff, thanks! For future reference, it's best to put new strings at the end of strings.xml to make things easier for translators (but I'll fix that now :))

PR #210, opened 2015-04-05
Handle redd.it links in comments. Fixes #166

### LINKED ISSUE #166: [Feature Request] support for reddit shortlinks in app
So right now when you open a reddit short link it opens it in a web view instead inside the APP.

Would be a nice feature

ajs124:
It seems like shortlinks don't take parmeters like context and so on. In case they do, they are not handled at the moment.


QuantumBadger:
Nice work, thanks!

PR #211, opened 2015-04-12
Reset image touch state on screen rotation

ThomasKomarnicki:
I noticed if I was zooming or dragging an image, and my screen rotated, the touch controls would work incorrectly. To replicate: open an image, hold a finger (or 2) on the screen and rotate your device. I attempted to fix this buy resetting the touch state on a configuration change. Thanks!


QuantumBadger:
I just tried this and can reproduce the issue. Thanks for the fix!

PR #219, opened 2015-05-28
Fixes #217 Collapsing big comments

### LINKED ISSUE #217: Collapsing big comments
When collapsing big comments you have to scroll back up to where you were before. So instead can we have it to automatically scroll up to the last collapsed comment when collapsing?

chaitan94:
I think this should fix #217


QuantumBadger:
Fantastic stuff, thanks!

Issue #159, opened 2014-08-07
[Feature Request] Option to open the front page by default instead of the main menu


evaryont:
I like going the front page. The subreddits I've picked are awesome/amazing. I find it a tad bit annoying and extra work to have to click on "Front page" every time I open the app. (I tend to close my apps on my phone completely rather than leave them running in the background.) An option to jump to the front page whenever the application starts would be much appreciated!

PR #220, opened 2015-05-28
Implemented #159: Option to open the front page by default instead of the main menu

chaitan94:


QuantumBadger:
Thanks!

As it stands this has a bug:

1) User opens app
2) App opens MainActivity, which immediate opens the front page
3) Android destroys the MainActivity because it's in the background
4) The user presses back to get to the MainActivity
5) MainActivity.onCreate() is called again, so the user is sent straight back to the front page, and can't access the main menu.

You can test this using the "Don't keep activities" setting in Developer Options. I assume it can be fixed using the saved instance state to detect whether the activity is being opened for the first time.

chaitan94:
You're right. I've added a check on savedInstanceState to see if the activity is being opened for the first time.

QuantumBadger:
Excellent, thanks!

Issue #221, opened 2015-06-01
Links to subreddits in a sidebar don't work


travisfw:
I viewed the sidebar of r/DataIsBeautiful and clicked on a linked subreddit r/Datasets and got a white screen with just the RedReader title: "Sidebar: /r/dataisbeautiful/"
RedReader version: 1.8.7.8
Android version: 4.4.2

chaitan94:
I think the issue is because of the relative link, after adding a base URL to reddit.com, the links seem to work. But still they work like regular external links i.e. they don't open in-app by default.

travisfw:
Probably true @chaitan94. I expect a solution probably entails rewriting the html.

QuantumBadger:
To fix this, some logic needs to be added here:

https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/fragments/WebViewFragment.java#L176

If we're showing a sidebar, we should prepend "https://reddit.com" to any URLs which start with "/".

For that we need an extra parameter here:

https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/fragments/WebViewFragment.java#L73

To say whether or not we're showing a sidebar.

chaitan94:
Are we keeping track of currentUrl?
If so, how about this: In https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/fragments/WebViewFragment.java#L176
If the clicked link's url start with "/", we can get the host name from currentUrl and prepend it to url before opening it.

Something like this:
```java
if(RedditURLParser.parse(Uri.parse(url)) != null) {
    String targetUrl = url;
    if(url.length() > 0 && url.startsWith("/")) {
        Uri curUri = Uri.parse(currentUrl);
        if (curUri != null)
            targetUrl = curUri.getScheme() + "://" + curUri.getHost() + url;
    }
    LinkHandler.onLinkClicked(getSupportActivity(), targetUrl, false);
} else {
    webView.loadUrl(url);
    currentUrl = url;
}
```

Cguy7777:
@QuantumBadger I think this was fixed at some point. I just double-checked on v1.15, and subreddit links in sidebars seem to work fine now.

PR #222, opened 2015-06-02
Issue #221: Links to subreddits in a sidebar don't work

chaitan94:
Added base url for the relative uris


QuantumBadger:
It looks like this would break non-reddit websites, which also use the WebViewFragment.

Issue #214, opened 2015-04-18
Lock orientation


ToNIX-:
Adding an option to lock the orientation would be a nice feature. This would prevent the phone from switching to landscape when tilting it.

Thanks.

chaitan94:
I think we have two options to implement this
1) Do a `setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_PORTRAIT);` in every activity class, whenever the preference is set.
2) Make a BaseActivity class, add it there and make all activities extend from it.
I guess second option is better as any future changes will be easier. I can send a pull request if you are fine with it.

Currently the Activities which extend from org.holoeverywhere.app.Activity are:
HtmlViewActivity
BugReportActivity
PostSubmitActivity
CommentEditActivity
InboxListingActivity
ImageViewActivity
OAuthLoginActivity
CaptchaActivity
RefreshableActivity
CommentReplyActivity
Additionally SettingsActivity extends from org.holoeverywhere.preference.PreferenceActivity

ToNIX-:
Good idea. Quickpic uses this, independently from the phone settings.

Change Screen Orientation
- By system setting
- By orientation sensor

QuantumBadger:
Added in v1.9.0 thanks to @chaitan94 :)

PR #227, opened 2015-06-20
Added ability to set preference for screen orientation throughout the app

chaitan94:
Issue #214 

User can set his preference for the screen orientation for the whole app.


QuantumBadger:
Looks good -- thanks!

PR #229, opened 2015-07-29
Switch over to Gradle

mvdan:
I moved the `res` folder to where it's stored in android Gradle projects. I can redo the PR keeping it in the old location, but in the long run you want to move it there.

I just set up the project and added the exact same deps that maven had.

PR #232, opened 2015-08-06
Random Subreddit

krisdb:
Wrote this for myself, not sure if you want to incorporate it.  I tried to resuse the existing json parser, but had a hard time following it, so wrote my own.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -151,7 +151,9 @@
     <string name="mainmenu_frontpage">Front Page</string>
     <string name="mainmenu_all">All Subreddits</string>
     <string name="mainmenu_custom">Custom Subreddit</string>
+    <string name="mainmenu_random">Random Subreddit</string>
     <string name="mainmenu_custom_invalid_name">Invalid subreddit name.</string>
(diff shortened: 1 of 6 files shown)


mvdan:
Looks like you branched off before #229 was merged, so there are conflicts. Could you rebase on top of the current master?

mvdan:
Reping @krisdb

QuantumBadger:
Before this is merged in, it should ideally:

* Make use of the CacheManager and its JSON parser, rather than manually making an HTTP request. This means it'll always respect reddit's API limits, and will make the request through the oAuth API.
* Include an option to hide the "random" option from the main screen of the app (with the option hidden by default)
* Detect attempts to go to /r/random via links (and via the "custom subreddit" option), and handle them correctly

If nobody else volunteers, I'll probably make the above changes myself for v1.9.1.

Zelpa:
so is this still on the radar of things to be fixed in the future?

QuantumBadger:
Closing this PR request as the code has significantly diverged, and I think there are better approaches to fixing this.

PR #233, opened 2015-08-08
Add intent filter support for reddit's mobile site

Septolum:


mvdan:
LGTM :)

ajs124:
Seems fine. Maybe we could add the old pay. ssl workaround while we're at it. I think it doesn't work yet, either.

QuantumBadger:
Merged, thanks!

> Maybe we could add the old pay. ssl workaround while we're at it. I think it doesn't work yet, either.

Will do.

PR #256, opened 2015-12-07
Updated Spanish translations <!-- 2015-12-06 -->

moshpirit:
I have a doubt: until now we could send as much submissions as we wanted? or the news is just the message shown. Are still going to appear the CAPTCHAs?


QuantumBadger:
Fantastic, thank you!

> until now we could send as much submissions as we wanted? or the news is just the message shown

The only difference is the message is now shown (before it would silently fail).

> Are still going to appear the CAPTCHAs?

Yes.

moshpirit:
OK, wonderful work then :)

Issue #24, opened 2013-04-09
Filter subreddits from r/all


pwcazenave:
Along the lines of the filtering capabilities in RES, the ability to omit specific subreddits or posts containing certain keywords from r/all would be particularly welcome. None of the other Android clients that I've come across have this functionality.

ishults:
I've taken a stab at adding some pretty basic functionality. It needs a unified view still to view/edit the blocked subreddits, but in the meantime it is possible to block subreddits from /r/all.

inl-pd-autotest:
Just block the subreddits you hate and they won't be shown on r/all, problem solved.
This requires an account though, but it's not like a hard thing to do.

ishults:
Reddit's built-in functionality did not exist when this feature request was added.  Also, it has a limit of 50 subreddits.

PR #258, opened 2015-12-20
#24 Add subreddit blocking functionality

ishults:
Still needs a unified view (in settings?) to add/remove/edit subreddits.  But the ability to filter on a per-subreddit level is available.


QuantumBadger:
Absolutely brilliant work, thank you!

I've taken a read through and it all looks good. My only (minor) gripe is that the indentation is a bit dodgy in a few places (tabs vs spaces), but it's already a bit inconsistent so don't worry about it :)

ishults:
Dang it, I know, sorry.  I tried to do my best, but like you said, some files are different from others, so it was hard to catch at times.  Thanks for the merge!

QuantumBadger:
No problem, thanks again for your hard work!

PR #269, opened 2016-01-03
Image size during image loading and Preference to skip gallery list view

jacobspeicher:
Display the full image size and the progress on the bottom of the screen under the circular progress bar.

Also added preference to skip gallery list view and go directly to the first image, then close the album with a single tap or view the gallery list by pressing the back button.


src/main/java/org/quantumbadger/redreader/common/LinkHandler.java
```diff
@@ -109,6 +111,8 @@ public void run() {
 			}
 
 			activity.startActivity(intent);
+			if(PrefsUtility.pref_behaviour_skipgallerylist(activity, PreferenceManager.getDefaultSharedPreferences(activity)))
+				activity.finish();
```
QuantumBadger: I think it would be better if this was done in the `AlbumListingActivity`.

src/main/java/org/quantumbadger/redreader/activities/AlbumListingActivity.java
```diff
@@ -136,12 +138,21 @@ public void onNotAnImage() {
...
-				AndroidApi.UI_THREAD_HANDLER.post(new Runnable() {
+				if(PrefsUtility.pref_behaviour_skipgallerylist(AlbumListingActivity.this, PreferenceManager.getDefaultSharedPreferences(AlbumListingActivity.this)) && !intent.hasExtra("viewList")) {
+					LinkHandler.onLinkClicked(
+							AlbumListingActivity.this,
+							info.images.get(0).urlOriginal,
+							false,
+							null,
+							info,
+							0);
+					General.quickToast(AlbumListingActivity.this, "Image 1/" + String.valueOf(info.images.size()), Toast.LENGTH_SHORT);
```
QuantumBadger: The string `"Image 1/` should be defined in `strings.xml`.

QuantumBadger:
Brilliant, thanks! I've added a couple of minor comments to the code.

Would it be possible to split this up into two pull requests? (one for the loading text and one for the list skipping) If it's too much hassle then don't worry about it (I assume the way to do it is to have two separate branches, and submit a PR for each).

src/main/java/org/quantumbadger/redreader/activities/AlbumListingActivity.java
```diff
@@ -136,12 +138,21 @@ public void onNotAnImage() {
 			@Override
 			public void onSuccess(final ImgurAPI.AlbumInfo info) {
 				Log.i("AlbumListingActivity", "Got album, " + info.images.size() + " image(s)");
-
-				AndroidApi.UI_THREAD_HANDLER.post(new Runnable() {
+				if(PrefsUtility.pref_behaviour_skipgallerylist(AlbumListingActivity.this, PreferenceManager.getDefaultSharedPreferences(AlbumListingActivity.this)) && !intent.hasExtra("viewList")) {
+					LinkHandler.onLinkClicked(
```
QuantumBadger: Actually I think this could cause a bug -- if the activity is stopped while the album is loaded, the image will be shown to the user anyway.

paetramon:
I moved the activity.finish(); code into the AlbumLisitingActivity and "Image 1/" into the strings.xml. I also moved the code to display the image instead of the list to after the list gets loaded, which hopefully fixes the possible bug you mentioned in one of your comments?

Also, I think trying to split it into multiple pull requests would be a bit of a process splitting it into multiple branches and such. I'll keep it in mind for future pull requests I make though!

QuantumBadger:
Thanks! I don't think the bug is fixed though -- the `onLinkClicked()` call was originally in the right place (on the UI thread), but it should only be called if the `onDestroy()` callback hasn't been received for the activity.

It should be sufficient to add a member variable `mActivityStopped` which you set to true in `onDestroy()`, and check before calling `onLinkClicked()`/`finish()`/showing the toast.

QuantumBadger:
Closing this PR as the code has diverged, and there are unresolved race conditions.

Issue #52, opened 2013-06-12
[feature] inbox - show only new


grrrrr:
Feature Request:

Have an option to load / display only new messages in the inbox.  This would allow user to not worry about if they are re-read old messages and maybe reply again by mistake.

QuantumBadger:
Will be in v1.9.3 :)

PR #273, opened 2016-01-25
Added option to only view unread messages in inbox.

H4rryA:
Implemented feature requested in issues #52. Added checkable item in inbox options menu to switch between viewing the inbox and viewing only unread messages.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/InboxListingActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/InboxListingActivity.java
@@ -60,8 +60,10 @@ import java.util.UUID;
 public final class InboxListingActivity extends BaseActivity {
 
 	private static final int OPTIONS_MENU_MARK_ALL_AS_READ = 0;
+	private static final int OPTIONS_MENU_SHOW_UNREAD_ONLY = 1;
 
 	private InboxListingAdapter adapter;
+    private SharedPreferences.Editor editor;
 
 	private LoadingView loadingView;
 	private LinearLayout notifications;
@@ -71,6 +73,7 @@ public final class InboxListingActivity extends BaseActivity {
 	private EnumSet<PrefsUtility.AppearanceCommentHeaderItems> headerItems;
 
 	private boolean isModmail = false;
+	private boolean onlyUnread;
 
 	private final Handler itemHandler = new Handler(Looper.getMainLooper()) {
 		@Override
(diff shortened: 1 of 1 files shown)


src/main/java/org/quantumbadger/redreader/activities/InboxListingActivity.java
```diff
@@ -295,6 +303,11 @@ public void onBackPressed() {
 	@Override
 	public boolean onCreateOptionsMenu(final Menu menu) {
 		menu.add(0, OPTIONS_MENU_MARK_ALL_AS_READ, 0, R.string.mark_all_as_read);
+		menu.add(0, OPTIONS_MENU_SHOW_UNREAD_ONLY, 1, "Unread Messages Only");
```
QuantumBadger: This string should be moved to `strings.xml`.

src/main/java/org/quantumbadger/redreader/activities/InboxListingActivity.java
```diff
@@ -342,6 +354,19 @@ public void run() {
...
+					item.setChecked(true);
+					editor.putBoolean("onlyUnread", true);
+					onlyUnread = true;
+				} else {
+					item.setChecked(false);
+					editor.putBoolean("onlyUnread", false);
+					onlyUnread = false;
+				}
+				editor.commit();
+                makeFirstRequest(this);
```
QuantumBadger: Wonky indentation (in a few other places too)

src/main/java/org/quantumbadger/redreader/activities/InboxListingActivity.java
```diff
@@ -342,6 +354,19 @@ public void run() {
 						this);
 
 				return true;
+			case OPTIONS_MENU_SHOW_UNREAD_ONLY:
+				if (!item.isChecked()) {
+					item.setChecked(true);
+					editor.putBoolean("onlyUnread", true);
```
QuantumBadger: This string (`onlyUnread`) should be in `strings.xml` and ideally should have a wrapper in `PrefsUtility.java`. See this for an example:

https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java#L160

QuantumBadger:
Nice work, thanks! There are a few minor things that could be fixed -- I've added some comments. If you get round to making the changes that would be fantastic, if not I'll do them myself before release.

Issue #282, opened 2016-03-13
Overlay icons cause posts with no thumbnails to rearrange text and bump content


ishults:
It seems that with the recent improvement to add overlay icons to certain post actions (e.g. upvoting), that if a post does not have a thumbnail, adding the overlay causes the text to spill onto more lines, and bump other posts down.  GIF screenshot attached:
![content shift](https://cloud.githubusercontent.com/assets/7275442/13731482/c4f856a2-e937-11e5-8485-7a49b39dadb3.gif)

In rare cases (that I haven't been able to pin down), it's even possible for the text to align onto FEWER lines, cutting off some of the content.

I believe I have a potential fix, in which we basically just set an overlay in place of the thumbnail and give it a default image (otherwise it doesn't look as good on the "Night" theme; I've used the comment icon for now).  This has the added bonus of lining up all text regardless of the presence of a thumbnail.

![potential fix 1](https://cloud.githubusercontent.com/assets/7275442/13746167/012d2340-e9c0-11e5-91c7-0551c4efabcb.png)

I realize this is a minor issue, so feel free to disregard this ticket if the fix isn't wanted/needed.

QuantumBadger:
I see your point, and I agree the relayout of the text is a bit odd. On balance though I'd say I prefer the current appearance (mainly because of the duplicate icon).

ishults:
Sure, no problem.  I can close it.

PR #283, opened 2016-03-13
#282 Fix text alignment issue

ishults:


QuantumBadger:
Hmm, not sure about this one. I'm not too keen on having the same "comments" icon at both ends of the screen, or with the wasted space on thumbnail-free subreddits.

I'll apply it locally though and take a look myself :)

Issue #271, opened 2016-01-09
Posts not marked as read when using "view comments" swipe action


QuantumBadger:

ishults:
Took a stab at this one as well.

PR #284, opened 2016-03-13
#271 Potential fix for post not marked as read

ishults:


QuantumBadger:
Looks good, thanks!

philipwhiuk:
Does this mark comments read when you tap on them from the Inbox?

QuantumBadger:
> Does this mark comments read when you tap on them from the Inbox?

No, it's only for posts in the post list. Unfortunately there isn't currently a way in the app to mark individual inbox items as read (although there is a "mark _all_ items as read" menu option).

PR #291, opened 2016-03-29
Updates to README.md

DistantParadox:
- Added image preview
- Added RedReader logo
- Reordered and added features
- Added link to F-Droid Store

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,10 @@ License
 
 RedReader is licensed under the GPL, version 3. A copy of the license is included in [LICENSE.txt](LICENSE.txt).
 
-Bitcoin donations are accepted at the following address: `1874wapGxDo2vEp4avisda4gx3SCjsHCQJ`
+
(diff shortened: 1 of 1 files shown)


DistantParadox:
If the preview file could be replaced with the original (by replacing the link), then it will probably provide higher quality, and thus be enlargened.  This preview image was grabbed from the Google Play store screenshots, and should suffice for now.

QuantumBadger:
Thanks for the pull request! Unfortunately I think I'm going to turn this down for now, because:

* The app is about to get a Material Design update, and I'd rather not post any screenshots until that goes live
* The images have JPEG artifacts and are externally hosted

It's a good idea to add images though and I'll add some when MD is done. If you'd like to resubmit just the F-Droid link (with the middle R in RedReader capitalised ;)) that would be fine.

DistantParadox:
No problem! I'll make the fix.

PR #296, opened 2016-04-03
Experimental TOR support

SpotComms:
- It doesn't use the NetCipher HTTP connection methods, since you're already using OkHTTP.
- GfyCat seems to be blocking API access from TOR

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/BaseActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/BaseActivity.java
@@ -17,17 +17,23 @@
 
 package org.quantumbadger.redreader.activities;
 
+import android.app.AlertDialog;
+import android.content.DialogInterface;
 import android.content.SharedPreferences;
 import android.content.pm.ActivityInfo;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.support.v7.app.AppCompatActivity;
 import org.quantumbadger.redreader.R;
+import org.quantumbadger.redreader.cache.CacheDownload;
 import org.quantumbadger.redreader.common.PrefsUtility;
+import org.quantumbadger.redreader.http.okhttp.OKHTTPBackend;
+import info.guardianproject.netcipher.proxy.OrbotHelper;
 
 public class BaseActivity extends AppCompatActivity implements SharedPreferences.OnSharedPreferenceChangeListener {
 
 	private SharedPreferences sharedPreferences;
+	private static boolean tor;
 
 	private static boolean closingAll = false;
 
@@ -42,6 +48,7 @@ public class BaseActivity extends AppCompatActivity implements SharedPreferences
 		sharedPreferences = PreferenceManager.getDefaultSharedPreferences(this);
 		sharedPreferences.registerOnSharedPreferenceChangeListener(this);
 		setOrientationFromPrefs();
+		setTorFromPrefs();
 		closeIfNecessary();
 	}
 
@@ -49,6 +56,7 @@ public class BaseActivity extends AppCompatActivity implements SharedPreferences
 	protected void onResume() {
 		super.onResume();
 		setOrientationFromPrefs();
+		setTorFromPrefs();
 		closeIfNecessary();
 	}
 
@@ -79,6 +87,35 @@ public class BaseActivity extends AppCompatActivity implements SharedPreferences
 			setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE);
 	}
 
+	private void setTorFromPrefs() {
+		tor = PrefsUtility.network_tor(this, sharedPreferences);
+		if (tor) {
+			if (OrbotHelper.isOrbotInstalled(this)){
+				if(!OrbotHelper.isOrbotRunning(this)) {
+					OrbotHelper.requestStartTor(this);
+				}
+			} else {
+				AlertDialog.Builder notInstalled = new AlertDialog.Builder(this);
+				notInstalled.setMessage(R.string.error_tor_not_installed);
+				notInstalled.setPositiveButton(R.string.dialog_yes, new DialogInterface.OnClickListener() {
+					public void onClick(DialogInterface dialog, int id) {
+						startActivity(OrbotHelper.getOrbotInstallIntent(getApplicationContext()));
+						dialog.dismiss();
+					}
+				});
+				notInstalled.setNegativeButton(R.string.dialog_no, new DialogInterface.OnClickListener() {
+					public void onClick(DialogInterface dialog, int id) {
+						dialog.cancel();
+					}
+				});
+				AlertDialog notInstalledAlert = notInstalled.create();
+				notInstalledAlert.show();
+			}
+		}
+		OKHTTPBackend.recreateHttpBackend();
+		CacheDownload.resetUserCredentialsOnNextRequest();
+	}
+
 	protected void onSharedPreferenceChangedInner(final SharedPreferences prefs, final String key) {
 		// Do nothing
 	}
@@ -93,6 +130,12 @@ public class BaseActivity extends AppCompatActivity implements SharedPreferences
 
 		} else if(key.equals(getString(R.string.pref_menus_optionsmenu_items_key))) {
 			invalidateOptionsMenu();
+		} else if(key.equals(R.string.pref_network_tor_key)) {
+			setTorFromPrefs();
 		}
+	}
+
+	public static boolean getTorStatus() {
+		return tor;
 	}
 }
--- a/src/main/java/org/quantumbadger/redreader/cache/CacheDownload.java
+++ b/src/main/java/org/quantumbadger/redreader/cache/CacheDownload.java
@@ -17,6 +17,7 @@
 
 package org.quantumbadger.redreader.cache;
 
+import org.quantumbadger.redreader.activities.BaseActivity;
 import org.quantumbadger.redreader.activities.BugReportActivity;
 import org.quantumbadger.redreader.common.PrioritisedCachedThreadPool;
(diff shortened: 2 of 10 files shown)


QuantumBadger:
Awesome, thanks! This looks good enough to merge in, although I have some comments/questions (which I'll add to the diff now).

src/main/java/org/quantumbadger/redreader/activities/BaseActivity.java
```diff
@@ -79,6 +84,18 @@ else if (orientation == PrefsUtility.ScreenOrientation.LANDSCAPE)
 			setRequestedOrientation(ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE);
 	}
 
+	private void setTorFromPrefs(boolean force) {
+		tor = PrefsUtility.network_tor(this, sharedPreferences);
+		if (tor) {
+			if (OrbotHelper.isOrbotInstalled(this)){
```
QuantumBadger: I think it would be better to check this in the preference fragment, and to not allow the user to select the preference if Orbot isn't installed (a dialog can be popped up to tell them why).

src/main/java/org/quantumbadger/redreader/activities/BaseActivity.java
```diff
@@ -79,6 +84,18 @@ else if (orientation == PrefsUtility.ScreenOrientation.LANDSCAPE)
...
+				OrbotHelper.requestStartTor(this);
+			} else {
+				OrbotHelper.getOrbotInstallIntent(this);
```
QuantumBadger: Does this line do anything? AFAICS getOrbotInstallIntent just returns an intent, and doesn't start the activity by itself.

src/main/java/org/quantumbadger/redreader/cache/CacheDownload.java
```diff
@@ -130,6 +136,9 @@ private void performDownload(final HTTPBackend.Request request) {
 		request.executeInThisThread(new HTTPBackend.Listener() {
 			@Override
 			public void onError(final RequestFailureType failureType, final Throwable exception, final Integer httpStatus) {
+				if(BaseActivity.getTorStatus()) {
```
QuantumBadger: It's possible to do this only if a reddit request fails by checking:

    initiator.queueType == CacheRequest.DownloadQueueType.REDDIT_API

src/main/java/org/quantumbadger/redreader/http/okhttp/OKHTTPBackend.java
```diff
@@ -31,11 +35,26 @@
...
+		}
+		return httpBackend;
+	}
 
 	public OKHTTPBackend() {
 
 		mClient = new OkHttpClient();
 
+		if(BaseActivity.getTorStatus()) {
+			Proxy tor = new Proxy(Proxy.Type.HTTP, new InetSocketAddress("127.0.0.1", 8118));
```
QuantumBadger: Maybe a silly question, but do HTTP proxies handle HTTPS requests? Either way it might be better to use SOCKS here (I think Orbot and OkHTTP both support it).

src/main/res/xml/prefs_network.xml
```diff
@@ -25,6 +25,10 @@
                         android:key="@string/pref_network_https_key"
                         android:defaultValue="true"/>
 
+    <CheckBoxPreference android:title="@string/pref_network_tor_title"
+        android:key="@string/pref_network_tor_key"
+        android:defaultValue="tor"/>
```
QuantumBadger: Should this be "false" rather than "tor"?

src/main/java/org/quantumbadger/redreader/views/WebViewFixed.java
```diff
@@ -37,4 +44,14 @@ public void onWindowFocusChanged(final boolean hasWindowFocus) {
 			Log.e("WebView", "WebView.onWindowFocusChanged", ex);
 		}
 	}
+
+	private void setProxy(Context context) {
+		if(BaseActivity.getTorStatus()) {
+			try {
+				WebkitProxy.setProxy("org.quantumbadger.redreader.RedReader", context.getApplicationContext(), this, "127.0.0.1", 8118);
```
QuantumBadger: This will break if the package name changes (e.g. in the Alpha version). Might be better to use `RedReader.getClass().getCanonicalName()`?

QuantumBadger: I think the return value should be checked here.

src/main/java/org/quantumbadger/redreader/http/okhttp/OKHTTPBackend.java
```diff
@@ -31,11 +35,26 @@
 public class OKHTTPBackend implements HTTPBackend {
 
 	private final OkHttpClient mClient;
+	private static HTTPBackend httpBackend;
+	public static boolean reset = false;
```
QuantumBadger: It might be better to move `reset` into `CacheDownload`, and add a static method called `CacheDownload.resetUserCredentialsOnNextRequest()` (so that the variable isn't public).

Also, requests are made from different threads, so this should probably be `volatile`.

src/main/java/org/quantumbadger/redreader/http/okhttp/OKHTTPBackend.java
```diff
@@ -31,11 +35,26 @@
...
+
+	public static HTTPBackend getHttpBackend(boolean force) {
```
QuantumBadger: Should be `synchronized`.

QuantumBadger: Also it might be better to split this into `getHttpBackend()` and `recreateHttpBackend()` to get rid of the parameter, but that's just nitpicking :)

QuantumBadger:
In addition to the code comments I've left, I don't think [`OAuthLoginActivity`](https://github.com/QuantumBadger/RedReader/blob/d31c838335e36cae28f5924f3810b6a2db515e61/src/main/java/org/quantumbadger/redreader/activities/OAuthLoginActivity.java) or the imgur album thumbnails are protected yet.

If you get a chance to address any of the comments, that would be fantastic -- otherwise, I'll probably do it myself before the v1.9.4 release. Thanks again!

mvdan:
Paging @eighthave who is the main author behind NetCipher and who IIRC is very interested in making it easier for apps to use it :)

mvdan:
Also this would be a huge plus to this Reddit client over the other clients available (besides being free software, of course)

SpotComms:
@QuantumBadger thanks for the detailed comments. I've addressed all of it except for moving setTorFromPrefs() to the preferences fragment. I also didn't change OkHTTP to SOCKS as it doesn't fails to perform DNS lookups as detailed in https://github.com/square/okhttp/issues/2315.

QuantumBadger:
Excellent, thanks!

Regarding this commit:

https://github.com/QuantumBadger/RedReader/pull/296/commits/9cee4122a43593c42de95750dabc973c91dbc0e7

I presume that didn't fix the 403 errors (due to [this later commit](https://github.com/QuantumBadger/RedReader/pull/296/commits/e10e7d00d023015d99abd11a16820e9f937c1c8d))? It would be good to revert it, as reddit might get angry if we abuse their API like that :)

SpotComms:
The later commit didn't fix it either. I honestly don't know why it keeps 403'ing. You'll be using it, it'll be fine then all of a sudden every single response after that will 403. Restarting the app and Orbot don't fix it, even when not logged in.

QuantumBadger:
That's very strange. Maybe it would be worth adding some logging to make sure the token is definitely getting refreshed? (although force closing the app should clear the token)

My first thought would have been that a particular set of Tor exit nodes are blocked by reddit, and that the 403 errors start occurring if Orbot switches to them. But in that case it's strange that restarting Orbot doesn't fix it (unless it coincidentally chose them again on startup).

When do the 403 errors stop? E.g. when the device is restarted? Or do they just stop after a while?

Also, do we get any payload/etc along with the 403 error? I might try and repro this myself later if I get some time.

SpotComms:
I've cleaned up all the commits and squashed them. I reverted the random device ID. The 403 issue is still happening and very mysterious. A reboot will almost always fix it, force killing Orbot and RedReader will sometimes fix it, and clearing app data for RedReader will sometimes fix it.

src/main/java/org/quantumbadger/redreader/http/okhttp/OKHTTPBackend.java
eighthave: If you can get it working with a SOCKS proxy, that's the better way to go because SOCKS is included directly in the Tor daemon, while the HTTP proxy is a separate daemon that wraps tor.  I have had little luck with SOCKS proxies in Android HTTP APIs.

eighthave: Oh, and, HTTP proxies handle HTTPS just fine. If the IP address is something other than 127.0.0.1, then this could be a security issue, since the browser would transmit in plain text to the proxy.

QuantumBadger:
> It might actually be a Orbot issue, its log keeps saying "Couldn't find Polipo process... retrying..."

I assume [Polipo](http://www.pps.univ-paris-diderot.fr/~jch/software/polipo/) is this HTTP proxy:

> If you can get it working with a SOCKS proxy, that's the better way to go because SOCKS is included directly in the Tor daemon, while the HTTP proxy is a separate daemon that wraps tor.

I'd find it odd if we got 403 errors due to it crashing though.

I'll merge this in now, and hopefully the cause of the 403 errors will become clear at some point. Thanks for your input everyone :)

PR #300, opened 2016-04-05
Allowing the webview to proceed when no change to the URL is needed

mariokmk:
You can simply return false from shouldOverrideUrlLoading() to allow the WebView to proceed loading a page when no change to the URL is required.

From the [shouldOverrideUrlLoading (WebView view, String url)](http://developer.android.com/reference/android/webkit/WebViewClient.html#shouldOverrideUrlLoading(android.webkit.WebView,%20java.lang.String)) docs:
"_If WebViewClient is provided, return true means the host application handles the url, while return false means the current WebView handles the url._"


QuantumBadger:
Thanks!

PR #302, opened 2016-04-07
Remember toggled comments outside of session

ishults:
A mechanism to remember whether comments have been toggled outside of the current session.  Useful for AMAs, AskReddit posts, etc. where the user may save the post and come back to it later. 

Has a key for the maximum value (currently set to 100), but in order to avoid adding even more settings for the user, this is currently non-configurable.

Borrows somewhat from the Reddit Enhancement Suite approach of just-in-time pruning (although it only removes one entry when the limit is reached).


QuantumBadger:
Thanks for the pull request! But this feature is already partially implemented in the `RedditChangeDataManagerVolatile` -- I'm planning to modify it (before the v1.9.4 release) to save all its contents to disk, which includes the collapsed status of comments. So unfortunately I'm going to have to reject this PR :(

ishults:
Ha no problem.  I started before I had merged latest and noticed that there was a lot more code in place than before.

PR #306, opened 2016-04-12
Remove spaces from custom subreddit input

ishults:
I wanted to add the ability to input subreddits into the "Custom Subreddit" dialog without having to delete spaces (my keyboard autocompletes and auto-adds them).  I didn't realize this feature was already supported though!  So this is just a change to the display when the user is inputting the subreddit, to ignore spaces.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java
@@ -26,10 +26,9 @@ import android.content.pm.PackageManager;
 import android.graphics.Color;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
-import android.view.Menu;
-import android.view.MenuItem;
-import android.view.View;
-import android.view.WindowManager;
+import android.text.Editable;
+import android.text.TextWatcher;
+import android.view.*;
 import android.widget.EditText;
 import android.widget.FrameLayout;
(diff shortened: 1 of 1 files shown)


philipwhiuk:
Spaces aren't allowed, but underscores are:

https://github.com/reddit/reddit/blob/862a7e0fc5142058a88044f9254a8ec5db977a77/r2/r2/models/subreddit.py#L112

ishults:
@philipwhiuk Correct, my change only affects spaces.  I don't think we should auto-correct spaces to underscores, as then it doesn't really resolve the issue; it just replaces the character to delete.  If I type "ask reddit", I mean "askreddit", not "ask_reddit".  Plus the removal (rather than the replacement) of spaces is inline with the existing logic. 

Or am I missing something?

philipwhiuk:
> Plus the removal (rather than the replacement) of spaces is inline with the existing logic.

What existing logic?

ishults:
@philipwhiuk https://github.com/ishults/RedReader/blob/remove_spaces_custom_subreddit/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java#L279

philipwhiuk:
Oh so we're replacing them when we actually do the API call anyway? Then definitely I agree this should be merged.

QuantumBadger:
Thanks!

PR #308, opened 2016-04-18
Migrate account dialog from ListView to RecyclerView

veyndan:
I've only migrated one ListView fragment/activity to RecyclerView. If this pull request is accepted, I'd be happy to work on the others.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/adapters/AccountListAdapter.java
+++ b/src/main/java/org/quantumbadger/redreader/adapters/AccountListAdapter.java
@@ -1,118 +1,178 @@
 /*******************************************************************************
  * This file is part of RedReader.
- *
+ * <p>
  * RedReader is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- *
+ * <p>
  * RedReader is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- *
+ * <p>
  * You should have received a copy of the GNU General Public License
  * along with RedReader.  If not, see <http://www.gnu.org/licenses/>.
  ******************************************************************************/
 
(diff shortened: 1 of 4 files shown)


QuantumBadger:
Nice work, thank you!

The main challenge will be replacing the (buggy) ListView in the PostListingFragment, because of the support for flinging posts left/right. This uses a [realistic physical simulation](https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/views/list/SwipableListItemView.java#L124) so that swiping feels as natural as possible, as opposed to how flinging usually works on Android (e.g. in Gmail) where the slightest horizontal movement can cause an item to fly off the screen. If you'd like to have a go at replacing this with a RecyclerView (while maintaining the existing behaviour), that would be awesome :)

veyndan:
No problem! I don't see the problem with the way that Android implements this by default, as I feel the difference isn't large enough to undertake so much work for it. Using an `ItemTouchHelper` for the RecyclerView may be possible, so I'll have a look into it when I have time. In the `ItemTouchHelper` you can always override the `getSwipeVelocityThreshold` and up it so not slight movements will cause it to fly off the screen. In my opinion, delegating this kind of behaviour to the Android support libraries will reduce maintenance costs. Please let me know your thoughts on the simulation and whether this implementation seems OK.

QuantumBadger:
My main concerns with the Android implementation are:

* If you swipe a certain distance, come to a stop so there's no velocity, and then let go, the item will still sometimes fly off the screen. It seems to be mostly based on distance rather than momentum, and is slightly unpredictable. Obviously there's no difference in functionality, but it just feels "jankier" (to use Google's terminology :))
* With the existing implementation, posts in the post list don't fly off the screen at all when you fling them -- they just trigger actions once they go past a certain distance (e.g. upvoting) and then smoothly return to their original position. I'm not sure that `ItemTouchHelper` supports that, although I may be wrong.

When I originally implemented swiping for the `ListView` I had to use an overlay to intercept the touch events and make a decision re: swiping vs scrolling. Since `RecyclerView` handles touch events more sensibly, I'm guessing all of the `ListOverlayView` code can be removed, and each item in the list can receive the swiping touch events directly. The amount of code needed for this (vs the overlay) should be significantly reduced :)

PR #310, opened 2016-04-22
Remove ListSectionHeader class

veyndan:
The adapter will most likely change soon to be migrated over to `RecyclerView`. The look of the section subheader has slightly changed to follow the [spec](https://www.google.com/design/spec/components/subheaders.html#subheaders-list-subheaders). The `ListItemView` will be removed later as well, so there will be a greater consistency between the section header and its elements.

As I'm refactoring the `MainMenuAdapter`, do you want me to add the subheader "Reddit" to the top, as it appears in code but is commented out, so I guess that you had the intention of adding that. Also a question on dividers: what do you think about having full bleed dividers above the subheaders (except the "Reddit" one if you want to add it), and getting rid of the dividers for the elements of the headers to reduce visual noise? Ref: [Under section Subheaders and dividers](https://www.google.com/design/spec/components/dividers.html#dividers-types-of-dividers).

Also, one more thing is what is the use of the `ContextMenuBuilder` in `ListItemView`? `setContextMenuBuilder` never seems to be set, and it's not clear what the intentions of this are.


QuantumBadger:
> The look of the section subheader has slightly changed to follow the spec

I've built this locally and (on my device at least) it looks a bit strange. Significantly preferred how it looked before ;)

> As I'm refactoring the MainMenuAdapter, do you want me to add the subheader "Reddit" to the top, as it appears in code but is commented out

I think I commented that out because I thought it looked better without. Feel free to remove it entirely if you want.

> Also a question on dividers:

I'd prefer to keep the dividers as they are now, although if you attached a screenshot of what you're proposing I'd certainly consider it.

> Also, one more thing is what is the use of the ContextMenuBuilder in ListItemView?

Good question... presumably it was used for something in the past, but isn't any more. Again, feel free to delete it :)

veyndan:
Yeah it will look strange for the moment, as I said the elements layout hasn't been updated as well (e.g. Front Page, My Profile). (This)[https://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0B6Okdz75tqQsVlJfa2txeUtLTFE/components_dividers_type6.png] is a screen shot of what I was thinking the dividers could look like. There is significantly less visual clutter, while reinforcing the separation of the sections. Thanks, I'll remove those two then.

QuantumBadger:
> Yeah it will look strange for the moment, as I said the elements layout hasn't been updated as well (e.g. Front Page, My Profile)

Okay. I'd rather not accept pull requests for unfinished design changes, but either way I think the existing layout is fine as it is :)

> This is a screen shot of what I was thinking the dividers could look like

Yep, I prefer how it currently looks.

* Having everything left-aligned with nothing at the right hand side would make the design look weird and unbalanced
* Removing the dividing lines would make it more difficult for right-handed users. At the moment, even though the text is on the left, the dividing lines show the full "tappable area" for each item, which means they don't have the stretch their thumb over to the left to tap it. (If that's unclear I can attach a diagram :))

The linked screenshot gets around the above points by having the right-aligned icons, but RR wouldn't have those.

veyndan:
OK, I'll try and convert it to a `RecyclerView` without the design changes. Pretty swamped with work atm, but when I get time I'll get round to it.

QuantumBadger:
Sounds good, thanks!

PR #315, opened 2016-05-06
Remove redundant parent views

veyndan:
Three of the xml layouts were wrapped in `LinearLayout`s, where they were only wrapping one view, hence making them redundant.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java
@@ -28,10 +28,13 @@ import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.text.Editable;
 import android.text.TextWatcher;
-import android.view.*;
+import android.view.Menu;
+import android.view.MenuItem;
+import android.view.View;
+import android.view.WindowManager;
 import android.widget.EditText;
 import android.widget.FrameLayout;
-import android.widget.LinearLayout;
+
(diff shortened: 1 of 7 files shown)


src/main/java/org/quantumbadger/redreader/activities/MainActivity.java
```diff
@@ -247,10 +260,7 @@ public void onSelected(final @MainMenuFragment.MainMenuAction int type, final St
 			case MainMenuFragment.MENU_MENU_ACTION_CUSTOM: {
 
 				final AlertDialog.Builder alertBuilder = new AlertDialog.Builder(this);
-				final LinearLayout layout = (LinearLayout) getLayoutInflater().inflate(R.layout.dialog_editbox, null);
-				final EditText editText = (EditText)layout.findViewById(R.id.dialog_editbox_edittext);
-
-				editText.requestFocus();
```
QuantumBadger: Were these `requestFocus()` calls supposed to be removed?

veyndan: I set `<requestFocus />` via xml instead as every time this `EditText` was referenced, `requestFocus()` was being called, so doing it via xml is more explicit.

QuantumBadger: Awesome, thanks! I'll go ahead and merge this.

PR #316, opened 2016-05-07
String plurals and string format

veyndan:
Convert some of the strings to use string format in xml and use plurals in xml rather than via code. Though the legitimacy of the translation stays unchanged, people who wish to translate the app will have a significantly easier job. Also lint seems to help with plurals indicating which languages need more quantity strings for their language, e.g. Czech is said to require a quantity string for "few". I updated all the ones that I could find, but as strings.xml is quite large (as expected), I may have missed some when looking through.

It also appears that the commits from the previous PR is here. I pushed this onto a different branch, but I think it is because I worked on the views in the main branch. I'll remember to push both of these on separate branches next time. Sorry about that. At least it is better than the last PR haha


veyndan:
It appears that this branch has now got conflicts, probably due to the reason stated in the last paragraph of the previous post. Do you want me to close this and just send a new PR?

QuantumBadger:
It looks like this won't merge in cleanly because the first three commits (from the other PR) have been squashed into one by Github. I can fix it on my end if you want, or if you'd like to do it locally it's:

    git remote add quantumbadger https://github.com/QuantumBadger/RedReader.git
    git fetch quantumbadger
    git checkout string
    git rebase -i quantumbadger/master

While in the above interactive rebase, delete the first three commits, and then save (`ZZ` in vim).

    git push -f

Let me know if you'd like me to do it on my end :)

QuantumBadger:
Ah, just saw your comment. Either way is fine :)

veyndan:
I think that's it, and thanks for the help. I think I need to hone down on my git skills a bit more

QuantumBadger:
No problem -- all looks good now, thanks!

PR #317, opened 2016-05-07
Compound drawables

veyndan:
Instead of having a `LinearLayout` with an `ImageView` and `TextView` it can be replaced with just a `TextView` and a compound drawable. This then makes both files for single list with icon redundant as it can be replaced with single list.


QuantumBadger:
Looks good, thanks!

PR #318, opened 2016-05-07
Vectors instead of png for toolbar icons

veyndan:
Use AppCompat vector drawables instead of pngs as vectors have many benefits over pngs.


QuantumBadger:
This seems to have reduced the size of the toolbar icons, but it still looks okay. Thanks!

PR #319, opened 2016-05-08
Turn InboxListing into a RecyclerView

veyndan:


QuantumBadger:
I've tested this and there seem to be a few problems:

- No dividing lines between entries
- Entries don't occupy the full screen width (in cases where the text is shorter than the screen width). This causes link buttons to be narrower than the screen, and means that long-clicking on the empty right-hand part of the entry does nothing
- Long-clicking the entry header does nothing
- Entry headers are occasionally truncated in weird places. I noticed this when I was implementing the RecyclerView in the comment listing, and IIRC it was fixed by ensuring that all the entries were the full screen width.
- No "selector" highlight on tapping/long clicking (although I can't remember if that worked before)

I can upload a screenshot if any of that is unclear :)

veyndan:
Oh that's not good. 

> No dividing lines between entries

I must have forgotten to put the dividers back in, but can do that quite easily with an `ItemDecoration`. 

> Entries don't occupy the full screen width (in cases where the text is shorter than the screen width). This causes link buttons to be narrower than the screen, and means that long-clicking on the empty right-hand part of the entry does nothing

I'm looking at the previous version and it didn't highlight it there either. I can implement that as I probably just need to set 'match_parent' on the root view.

> Long-clicking the entry header does nothing

I'm not totally sure what you mean by the entry header. Please could you elaborate on that?

> Entry headers are occasionally truncated in weird places. I noticed this when I was implementing the RecyclerView in the comment listing, and IIRC it was fixed by ensuring that all the entries were the full screen width.

I don't seem to be getting this on my phone at all. Do you mind sending a screen shot of it?

> No "selector" highlight on tapping/long clicking (although I can't remember if that worked before)

You're right, it wasn't on the previous version either. I can easily implement this as well if you wish.

QuantumBadger:
By the "entry header" I mean the username/timestamp at the top of each entry.

> I must have forgotten to put the dividers back in, but can do that quite easily with an  ItemDecoration . 

When doing the comment listing I avoided ItemDecoration as it interacts weirdly with animations -- I ended up adding a 1px tall View to the top of each RedditCommentView:

https://github.com/QuantumBadger/redreader/blob/master/src/main/java/org/quantumbadger/redreader/views/RedditCommentView.java#L129

> You're right, it wasn't on the previous version either. I can easily implement this as well if you wish.

That would be awesome, thanks :)

> I don't seem to be getting this on my phone at all. Do you mind sending a screen shot of it?

Here's a screenshot -- see the second item down ("14 days, 1"):

screenshot_2016-05-10-20-58-11 out

veyndan:
That's pretty odd. I'll try and fix it when I have some time

PR #321, opened 2016-05-09
Convert CachedHeaderView into xml

veyndan:
Converting this into xml resulted in a layout with a view depth of 5, into a layout with a view depth of 1 (a single view).


QuantumBadger:
Nice, thanks!

PR #322, opened 2016-05-10
Convert LinkDetailsView into xml

veyndan:


QuantumBadger:
Sorry I didn't look at this sooner, I didn't get an email for some reason!

There seem to be some regressions though:

* Plain links show duplicate text for the button title/subtitle
* The globe image looks weird now
* The font size (or maybe the font itself) is different

**Before screenshot (current behaviour):**

pr322_before

**After screenshot:**

pr322_after

veyndan:
Sorry, didn't seem to notice the duplicate text. For the icons I just took the one from the material design website as the vector drawable is already created for it, but I guess I can revert it. I think I changed the font size, but definitely didn't change the typeface. I believe it was originally at 15sp but changed it to 14sp as 15 isn't a font size that Google recommends. I can always change it back. Haven't got much time over the next month or so, but can do it afterwards.

PR #323, opened 2016-05-12
Increase WebView anonymity when using Orbot

SpotComms:
This change makes every WebView have its cookies/cache/history/formdata cleared when created if the Orbot option is set for enhanced anonymity/security.


QuantumBadger:
Thanks!

PR #324, opened 2016-05-21
Make comment long-tap behavior dependent on comment tap behavior setting

mateoeh:
i.e. a long-tap will cause a comment to collapse if a normal tap is set to show the action menu, and vice versa. This way, tap and long-tap won't do the exact same thing when the tap action is set to 'Action Menu.'


QuantumBadger:
Looks good, thanks!

QuantumBadger:
I ended up making this a preference, and changing the defaults: https://github.com/QuantumBadger/RedReader/commit/15aa0c7ae27b9dd1b29f95cf33fd0d2f31edfcd6

PR #325, opened 2016-05-22
Version bumps and warning fixes

mvdan:


QuantumBadger:
Thanks! Before I merge this, would it be possible to undo the deletion of `icon_inv.png`? It's used as the icon for the [alpha version](https://redreader.org/alpha/).

At some point I'll modify the alpha building script to copy in `icon_inv.png` as needed, to save a few kB in the non-alpha apk :)

mvdan:
Ah, sure. One second.

mvdan:
Done.

QuantumBadger:
Awesome, thank you!

mvdan:
Your use of the "squash before merge" mechanism on github is interesting. I wouldn't use it as a rule though, as generally commits are split for a reason. For example, if any of the commits turn out to be troublesome, you won't be able to revert easily.

QuantumBadger:
> For example, if any of the commits turn out to be troublesome, you won't be able to revert easily.

Yep, agreed -- it also reduces the ability to use `git bisect` to figure out which change introduced an issue. But it's a trade-off between that and having a nice clean history without any diamonds/merge commits/etc. I'd significantly prefer it if GitHub added a "rebase without squashing" option :)

mvdan:
You can still do the regular rebase-merge without squashing via the shell. Check out the branch, rebase on top of master, merge. Don't know if GitHub will pick that up as merged though - I hope it would.

QuantumBadger:
I've done that in the past and IIRC GitHub didn't pick it up (so the commits weren't linked to the PR, and I had to close the PR manually). I suppose there's no way for them to tell, because there would be no merge commit, and the commit IDs would be different after the rebase. But it's strange that they have such poor support for such a common Git workflow.

PR #330, opened 2016-05-28
Update embedded Roboto font

### LINKED ISSUE #216: Remove embedded Roboto font
It seems that the app uses an outdated version of Roboto for some elements but uses the device's default font for others. The embedded fonts should be removed.
demo

mateoeh:
Fixes #216


QuantumBadger:
Thanks!

PR #334, opened 2016-06-04
Add Annotations

veyndan:
It's annoying that a lot of implemented classes are not annotated with the override annotation, so I did it for a couple of instances of it for the moment. I also added thread annotations and got rid of the comments of which thread it should run on, so IDE flagging can be enabled.


QuantumBadger:
Merged, thanks!

PR #335, opened 2016-06-09
Fixes #311 and #303: Blank title notifications, Disable commenting, voting and editing on an archived post

### LINKED ISSUE #311: Blank title notifications
When the user attempts to make a post with a blank title or a title consisting only of spaces, a "Post submitted" is generated even though no post is actually submitted.  This differs from the web client behavior which notifies the user that they need to choose a title.

chaitan94:
Shows a small Toast message to fill out the empty inputs, along with focusing onto them.
I believe it fixes #311

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/PostSubmitActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/PostSubmitActivity.java
@@ -141,11 +141,11 @@ public class PostSubmitActivity extends BaseActivity {
 		final Object selected = typeSpinner.getSelectedItem();
 
 		if(selected.equals("Link") || selected.equals("Upload to Imgur")) {
-			textEdit.setHint("URL"); // TODO string
+			textEdit.setHint(R.string.submit_post_url_hint);
 			textEdit.setInputType(InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_URI);
 			textEdit.setSingleLine(true);
 		} else if(selected.equals("Self")) {
-			textEdit.setHint("Self Text"); // TODO string
+			textEdit.setHint(R.string.submit_post_self_text_hint);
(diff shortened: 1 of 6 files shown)


chaitan94:
I'm sorry, I seem to have pushed both fixes in a single pull request :sweat: 

Since the fix for #311 is quite small I guess this might be fine. However, let me know if should rebase the commit and send in as separate pull requests.

src/main/java/org/quantumbadger/redreader/activities/PostSubmitActivity.java
```diff
@@ -187,10 +188,23 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
 		} else if(item.getTitle().equals(getString(R.string.comment_reply_send))) {
-			final Intent captchaIntent = new Intent(this, CaptchaActivity.class);
-			captchaIntent.putExtra("username", (String)usernameSpinner.getSelectedItem());
-			startActivityForResult(captchaIntent, REQUEST_CAPTCHA);
-
+			String subreddit = subredditEdit.getText().toString();
+			String postTitle = titleEdit.getText().toString();
+			String text = textEdit.getText().toString();
+			if (subreddit.isEmpty()) {
+				Toast.makeText(this, "Specify subreddit", Toast.LENGTH_SHORT).show();
```
QuantumBadger: This string (and the ones on line 198 and 201) should be declared at the end of strings.xml (to allow for translation).

QuantumBadger:
Mostly looks good, thanks! If you could move those strings into strings.xml, I'll get this merged :)

chaitan94:
I've move them.
Also while I'm on it, do you think I should move the strings on lines 144 and 148 as well? That would take care of the string on line 200 as well.

QuantumBadger:
That would be good, thanks!

chaitan94:
Done!

QuantumBadger:
Awesome, thank you!

PR #340, opened 2016-07-01
Tidy up Toolbar creation

veyndan:
I removed `OptionsMenuUtility.fixActionBar` as the method name, `fixActionBar` is nonsensical and using this method added an extra layer of complexity whereas using `setTitle` on the support action bar is shorter and already easy to understand.

Also regarding the `rrActionBarCol`, it is generally assumed that it is the primary colour of the app, so setting an extra attribute seemed a bit verbose.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/AlbumListingActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/AlbumListingActivity.java
@@ -53,12 +53,11 @@ public class AlbumListingActivity extends BaseActivity {
 
 		super.onCreate(savedInstanceState);
 
-		OptionsMenuUtility.fixActionBar(AlbumListingActivity.this, getString(R.string.imgur_album));
-
 		ActionBar ab = getSupportActionBar();
 		if (ab != null) {
 			ab.setHomeButtonEnabled(true);
 			ab.setDisplayHomeAsUpEnabled(true);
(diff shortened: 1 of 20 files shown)


QuantumBadger:
`fixActionBar` and `rrActionBarCol` had other uses in the past, but I agree they're unnecessary now.

Please could you revert f0de77b and 2134e46 -- I deliberately fixed the text size/actionbar height because of the horrible/inconsistent job Android does when the screen is rotated.

I'm on the fence about the null checks which have been added! I prefer things to crash rather than to silently do the wrong thing, and if `getSupportActionBar()` returns null after some future change, I'd like to be notified rather than just have some buttons be missing somewhere.

Utlimately this should be two PRs:

* One to remove the action bar colour attribute
* One to replace `fixActionBar` with straightforward calls to `setTitle`

veyndan:
OK I'll revert the `null` checks, but I don't understand what you mean by the inconsistent `toolbar` height and text size. Do you mind giving an example?

QuantumBadger:
The action bar/toolbar is sized differently depending on whether the device is in portrait or landscape. Unfortunately it decides on the size when the activity is created, and doesn't re-decide after rotating (since RedReader's activities make use of the `configChanges` property).

Even if a fix were found for this, I think the default is too small in landscape, and too large in portrait, so I prefer the way it looks now :)

PR #352, opened 2016-07-15
Fix image save for some URLs

0xKD:
`.getPath()` will return`/564x/57/56/e1/5756e17a05d3aeab42c32ebd99898866.jpg` for a URI like `https://s-media-cache-ak0.pinimg.com/564x/57/56/e1/5756e17a05d3aeab42c32ebd99898866.jpga`. (From [here](https://www.reddit.com/r/funny/comments/4sv4ez/aint_no_party_like/?))

This is an invalid filename (containing multiple `/`) and fails at https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/common/General.java#L114

This will remove all instances of `/` (File separator) from the output of `.getPath()` when being used as a filename.

Test against: https://www.reddit.com/r/funny/comments/4sv4ez/aint_no_party_like/


QuantumBadger:
Looks good, thanks!

PR #355, opened 2016-07-18
Add simple search functionality for comments

cpalasanu:
Hey QuantumBadger! 

I saw the issue about searching comments and gave it a shot.

 * Because of the "Load more" comments this patch doesn't search through all the comments but only the visible ones
 * Search is also available in the MoreCommentsListingActivity
 * Not sure how to handle dual-pane mode
 * Modified the SpecificCommentThreadView to notify the user when he is viewing search results

I'm not sure if this is what you had in mind. Let me know if I'm close.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
@@ -50,6 +50,7 @@ public final class OptionsMenuUtility {
 		CLOSE_ALL,
 		SUBMIT_POST,
 		SEARCH,
+		SEARCH_COMMENTS,
 		REFRESH_SUBREDDITS,
 		REFRESH_POSTS,
 		REFRESH_COMMENTS,
@@ -149,6 +150,7 @@ public final class OptionsMenuUtility {
 						add(activity, pastMenu, Option.PAST_COMMENTS, true);
 					}
 				}
+				if(optionsMenuItemsPrefs.contains(OptionsMenuItemsPref.SEARCH)) add(activity, menu, Option.SEARCH_COMMENTS, false);
 
 			} else if(postsVisible) {
 				if(postsSortable) {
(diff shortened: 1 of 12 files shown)


QuantumBadger:
Thanks! I've taken a look and this is pretty good!

It would be great if this also worked with the two-pane view (MainActivity). Probably the best way to do it is to open up a new CommentListingActivity, with the search string added as one of the parameters of the Intent. This approach could be used for the single pane view as well (instead of setting a variable and refreshing). Please let me know if you'd like to have a go at doing this!

cpalasanu:
Sure! I'll give it a go this weekend.

QuantumBadger:
Awesome, thanks!

src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
```diff
@@ -50,6 +50,7 @@
 		CLOSE_ALL,
 		SUBMIT_POST,
 		SEARCH,
+		SEARCH_COMMENTS,
```
cpalasanu: To tell the difference between post search and comment search in dual pane mode

src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
```diff
@@ -148,6 +150,7 @@
 						add(activity, pastMenu, Option.PAST_COMMENTS, true);
 					}
 				}
+				if(optionsMenuItemsPrefs.contains(OptionsMenuItemsPref.SEARCH)) add(activity, menu, Option.SEARCH_COMMENTS, false);
```
cpalasanu: If user disables search from the settings screen, don't show comment search either

src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
```diff
@@ -328,16 +331,37 @@ public boolean onMenuItemClick(final MenuItem item) {
 				break;
 
 			case SEARCH:
-				menu.add(activity.getString(R.string.action_search))
+				menu.add(Menu.NONE, Menu.NONE, 1, activity.getString(R.string.action_search))
```
cpalasanu: Added the order so that "Search" and "Search Comments" are next to each other

cpalasanu:
For dual pane mode I added a Menu item "Search Comments". For normal screens I left the regular "Search" item because it is obvious for the user what he is searching.

Fixed a case sensitivity issue with the search.

QuantumBadger:
Excellent work, thank you! I'll merge this in.

QuantumBadger:
Part of [Alpha 52](https://redreader.org/alpha/).

PR #367, opened 2016-08-06
Share post comments as permalink

fhtagn:
Hey!

I wanted to have the option of sharing a link using the "permalink" option, (the url in "X comments")

capture)

By reading [this post](https://www.reddit.com/r/redditdev/comments/1k6lz5/how_can_i_build_comment_permalink_from_ids/cblwrrb), I understand it is the "slug" part I am interested in, as it can be anything, but in particular it is already available from the API as a snippet of the post title, and placed in <code>RedditPost.java#permalink</code>

Added it as an option under Behaviour -> Posts. I understand the short link is better for sharing and saving space, so I made it an option, and by default it is off.

The feature was born out of a discussion with a friend: I send too many links and with the long url he is able to filter uninteresting ones, or those he has already seen, without the need to click.

Thank you for the client!

oajoul


QuantumBadger:
Looks good, thank you!

Would it be possible to move the new strings down to the bottom of `strings.xml`? They're ordered by date to make it easy for translators to see which ones are new.

fhtagn:
Ok, done, and added the date.

QuantumBadger:
Excellent, thanks! I'll merge this now.

QuantumBadger:
It looks like something weird has happened to the file permissions (they've gone from 644 to 755).  Not sure why :)

QuantumBadger:
Fixed here: https://github.com/QuantumBadger/RedReader/commit/b84956c0ce339a1fd21d32a82b6a6345e843d803

fhtagn:
Sorry. Currently in Windows, [seems to be part of the problem](https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/0EdNev3NNsw).

Glad it is done!

PR #370, opened 2016-08-07
Add preference for configuring cache location

### LINKED ISSUE #353: Add option to select cache storage location
My device has a small internal storage and a much larger external storage card.  The default "external" storage is the emulated internal storage though.  I should be able to choose from the full list of external and internal app cache directories to make sure I'm using the right one.

This is implemented by Google play music, [OsmAnd](http://osmand.net), and [AntennaPod](http://antennapod.org), if you'd like to see an example of it.  The latter two are open-source.

ShadowNinja:
This allows you to chose from the internal storage and all external storage locations.
Changing the location will not migrate existing cache files, but they will still be
loaded from the old location (and will exentually be pruned).

The AlertDialog-related code is borrowed from [AntennaPod](http://antennapod.org/), which is MIT licensed.

Note: Your XML files (or at least strings.xml) seem to use a mix of tabs and spaces.  The editorconfig suggests that you want tabs, but I set it to spaces in my code, because that was what the majority of the file used and I wanted it to blend in.  You may want to do a conversion to one of the two later though.

Fixes #353.

### DIFF
--- a/assets/changelog.txt
+++ b/assets/changelog.txt
@@ -2,6 +2,7 @@
 Multireddit viewing support
 Ability to view custom user pages
 Preference to use un-shortened link for "Share Comments" (thanks to fhtagn)
+Preference for cache storage location
 
 73/1.9.5.1
 Comment swipe actions
--- a/src/main/java/org/quantumbadger/redreader/cache/CacheManager.java
+++ b/src/main/java/org/quantumbadger/redreader/cache/CacheManager.java
@@ -20,6 +20,7 @@ package org.quantumbadger.redreader.cache;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.net.Uri;
+import android.os.Build;
 import android.preference.PreferenceManager;
 import android.util.Log;
 import org.quantumbadger.redreader.account.RedditAccount;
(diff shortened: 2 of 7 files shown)


QuantumBadger:
Thanks, this looks great!

`WRITE_EXTERNAL_STORAGE` is also used for saving images, so I think it's necessary even on newer platforms.

Also, how gracefully does this handle the selected directory disappearing (e.g. removing the SD card)? It would be worth adding a new `RequestFailureType` (and associated error message) to specifically say that the selected cache dir is no longer available.

Generally this looks fine though and pretty much ready to merge.

ShadowNinja:
You're right about the permission, I've removed that part of the patch.

It should simply fail to write new cache entries, and probably show an error message if it has a cache storage fail message.  The user will have to select a new storage directory from the settings.  I could have it automatically select a new storage dir if the location becomes unwritable, but I don't like the idea of settings just being changed without your knowledge.  I think it would be best if it showed an error message with on option to change the cache storage location, but that's a lot more work :-P.

QuantumBadger:
Okay thanks, I'll get this merged in (and maybe look at adding that error message myself :))

QuantumBadger:
Error message added (and a crash bug fixed) here: https://github.com/QuantumBadger/RedReader/commit/de77c825558435a986f2ff5b041e227d32c6be94

Issue #372, opened 2016-08-12
Pinned subreddits are sorted by date added


japanesephundroid:
Is this intended behaviour? Sort by name seems like a better option.

QuantumBadger:
PR merged in.

japanesephundroid:
Is the fix in **Alpha 71*? I tested it and the bug is still there.

QuantumBadger:
Nope, but I've just triggered Alpha 72, which has it in :)

QuantumBadger:
And Alpha 73 will be here in a few moments with a bugfix :P

japanesephundroid:
Thanks, Alpha 73 works for me :)

Symbian9:
I still has issue with sorting of [pinned subreddits](https://github.com/QuantumBadger/RedReader/issues/568) in alphabetical order in *RedReader* `v1.9.10`.

It still sorted in by date pinned.

PR #374, opened 2016-08-12
#372-sort pinned subbreddits based on name

chippen:
Second try, now with correct commit details.


QuantumBadger:
Looks fine thanks, I'll merge this in!

If you'd be interested in adding a preference for this (a ListPreference, with either "Alphabetical" or "By Date", and with an enum in PrefsUtility.java), here's an example of how to do it:

https://github.com/QuantumBadger/RedReader/commit/84d59cd

chippen:
I will look into it.

PR #376, opened 2016-08-15
#372 sort pinned subreddits

chippen:
This is almost done. I had two issues.
1. I did not understand how to make it refresh when the setting was changed. Now you have to do a manual refresh. (I solved the refresh part in a1e36d9)
2. I don't understand why you don't have to use android:summary="%s" (I had to)

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
+++ b/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
@@ -79,7 +79,8 @@ public final class PrefsUtility {
 				|| key.equals(context.getString(R.string.pref_behaviour_fling_post_right_key))
 				|| key.equals(context.getString(R.string.pref_behaviour_nsfw_key))
 				|| key.equals(context.getString(R.string.pref_behaviour_postcount_key))
-				|| key.equals(context.getString(R.string.pref_behaviour_comment_min_key));
+				|| key.equals(context.getString(R.string.pref_behaviour_comment_min_key))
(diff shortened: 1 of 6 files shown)


src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
```diff
@@ -404,6 +405,14 @@ public static PostSort pref_behaviour_postsort(final Context context, final Shar
 		return PostCommentListingURL.Sort.valueOf(General.asciiUppercase(getString(R.string.pref_behaviour_commentsort_key, "best", context, sharedPreferences)));
 	}
 
+	public enum PinnedSubredditSort {
+		NAME, DATE
+	}
+
+	public static PinnedSubredditSort perf_behaviour_pinned_subredditsort(final Context context, final SharedPreferences sharedPreferences) {
```
QuantumBadger: Minor typo (pref vs perf :P)

QuantumBadger:
This looks great, thanks! There's a minor typo which I've commented on, but other than that this looks ready to merge!

I think setting the summary to "%s" sets the text underneath the preference to its value (although there's a RR-specific mechanism for that here: https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/settings/SettingsFragment.java#L71)

chippen:
I fixed the typo and the summary

QuantumBadger:
Perfect, thank you!

PR #377, opened 2016-08-16
Add setting for only allowing NSFW content

chippen:
For fun I added a setting for filtering out non NSFW content.

But I had to clear my data in order to change the bool to a string for the setting. I do not know how to get around this, you might not want to merge this yet?

### DIFF
--- a/src/main/res/values/arrays.xml
+++ b/src/main/res/values/arrays.xml
@@ -574,4 +574,16 @@
 		<item>url</item>
 	</string-array>
 
+	<string-array name="pref_behaviour_nsfw">
+		<item>No NSFW Content</item>
+		<item>Show NSFW Content</item>
+		<item>Only NSFW Content</item>
+	</string-array>
+
+	<string-array name="pref_behaviour_nsfw_return">
+		<item>NO_NSFW</item>
+		<item>YES_NSFW</item>
+		<item>ONLY_NSFW</item>
+	</string-array>
+
 </resources>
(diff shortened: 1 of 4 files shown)


src/main/res/values/arrays.xml
```diff
@@ -574,4 +574,16 @@
 		<item>url</item>
 	</string-array>
 
+	<string-array name="pref_behaviour_nsfw">
+		<item>No NSFW Content</item>
```
QuantumBadger: These should be defined at the end of strings.xml to allow for translation :)

Also I'd change the option text to "Block", "Allow", and "Only show NSFW".

src/main/res/xml/prefs_behaviour.xml
```diff
@@ -130,9 +130,11 @@
 
     <PreferenceCategory android:title="@string/pref_behaviour_posts_header">
 
-        <CheckBoxPreference android:title="@string/pref_behaviour_nsfw_title"
-                            android:key="@string/pref_behaviour_nsfw_key"
-                            android:defaultValue="false"/>
+        <ListPreference android:title="@string/pref_behaviour_nsfw_title"
+						android:key="@string/pref_behaviour_nsfw_key"
```
QuantumBadger: When altering a preference, a different key should be used to the original (I'd go for `@string/pref_behaviour_nsfw_2_key` which should be added to strings.xml with the text `pref_behaviour_nsfw_2_key`).

QuantumBadger:
Thanks -- I've added a couple of comments, although in general it's hard to see a realistic use-case for this :P

> But I had to clear my data in order to change the bool to a string for the setting

This is because the same key was used for both the old and the new preference. The old value was still stored and was conflicting with the new preference type.

QuantumBadger:
Closing this for now due to lack of activity.

Issue #313, opened 2016-05-03
Add ability to view/edit currently blocked subreddits in settings


ishults:
As of right now you must go into the subreddit to block/unblock it and see its status.  Ideally, there would be a central list to add/remove/view/edit these.

PR #386, opened 2016-09-11
Display list of blocked subreddits on main menu

gstraube:
Hi, QuantumBadger!

This pull request is related to issue #313 (being able to view blocked subreddits). The blocked subreddits are listed in their own section on the main menu (akin to the pinned subreddits).

Concerning the removal of blocked subreddits: One way to implement this could be by left/right swiping the subreddit list item. The same approach could also be applied to unpinning or unsubscribing subreddits. What do you think?

Last but not least: I really enjoy using RedReader and very much appreciate the open source aspect and GPL licensing! Thank you for creating it!

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/RefreshableActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/RefreshableActivity.java
@@ -58,8 +58,11 @@ public abstract class RefreshableActivity extends BaseActivity {
 			return;
 		}
 
-		if(this instanceof MainActivity && key.equals(getString(R.string.pref_pinned_subreddits_key))) {
-			requestRefresh(RefreshableFragment.MAIN, false);
+		if(this instanceof MainActivity) {
+			if(key.equals(getString(R.string.pref_pinned_subreddits_key)) ||
+					key.equals(getString(R.string.pref_blocked_subreddits_key))) {
+				requestRefresh(RefreshableFragment.MAIN, false);
+			}
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


QuantumBadger:
Thank you! I've had a read through the code and this all looks good.

One quick request before merging: would it be possible to add a preference to hide this? I'm guessing some people will complain if their least favourite subreddits are permanently part of the menu :)

Swiping to remove would be quite easy to accidentally trigger, although I suppose it would be okay if there was an "undo" option (similar to e.g. Gmail). The simplest thing would be to add a long-click menu.

Again, thanks for submitting this -- hope to get it merged soon :)

philipwhiuk:
The original request was about putting it in settings. It definitely seems absurd to have it in the main menu.

gstraube:
Thank you for your feedback! You're right, it definitely makes sense to have an option to hide this (and to hide it by default). I'll get to it. Since there is currently no comparable functionality in the settings that deals with a list of subreddits, the main menu seemed to me like a reasonable place. By having it in the main menu, code that manipulates a list of subreddits can be re-used instead of re-implemented, e.g. when adding the functionality to remove a list item :-) (where a long-click menu is indeed a much better option).

QuantumBadger:
Sounds good, thanks! While the main menu isn't the ideal place for this, it's an improvement over the present behaviour (where they're completely inaccessible). Maybe in the future this could be moved to the settings (perhaps as a general subreddit management screen based on the main menu code), but for now this is fine.

gstraube:
I added the preference to show blocked subreddits in the main menu (disabled by default). Sorry for the delay!

QuantumBadger:
Excellent, thank you! I'll merge this in.

gstraube:
Thanks a lot!

PR #394, opened 2016-09-26
Allow navigation using the toolbar in 2 column mode

mariokmk:
Show the toolbar back button when reading comments in two column mode


QuantumBadger:
Thank you! Looks good to me.

PR #400, opened 2016-10-13
updates Spanish translation for 1.9.x

auroszx:
Hello. I saw there were some strings pending to be translated for the latest update, so I thought I'd give you a hand. Let me know if anything should be changed. Like the new features. Thank you.

PS: love the strings.xml layout, makes it easier to translate.


QuantumBadger:
Excellent stuff, thank you!

No need to translate language names (lang_nl, etc), but I'll fix that in a moment :)

PR #415, opened 2016-12-04
fix filename at General.java

### LINKED ISSUE #411: Add extension to reddituploads saved images
See: https://www.reddit.com/r/RedReader/comments/5ch5kq/bug_images_from_reddituploads_arent_saved_properly/

Bragado:
bug fix #411

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/General.java
+++ b/src/main/java/org/quantumbadger/redreader/common/General.java
@@ -401,7 +401,12 @@ public final class General {
 	private static final Pattern urlPattern = Pattern.compile("^(https?)://([^/]+)/+([^\\?#]+)((?:\\?[^#]+)?)((?:#.+)?)$");
 
 	public static String filenameFromString(String url) {
-		return uriFromString(url).getPath().replace(File.separator, "");
+		String filename = uriFromString(url).getPath().replace(File.separator, "");
+		String[] parts = filename.substring(1).split("\\.", 2);
+		if(parts.length < 2)
+			filename += ".jpg";
+
+		return filename;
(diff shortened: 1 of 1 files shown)


src/main/java/org/quantumbadger/redreader/common/General.java
```diff
@@ -401,7 +401,12 @@ public void onClick(DialogInterface dialog, int which) {
 	private static final Pattern urlPattern = Pattern.compile("^(https?)://([^/]+)/+([^\\?#]+)((?:\\?[^#]+)?)((?:#.+)?)$");
 
 	public static String filenameFromString(String url) {
-		return uriFromString(url).getPath().replace(File.separator, "");
+		String filename = uriFromString(url).getPath().replace(File.separator, "");
+		String[] parts = filename.substring(1).split("\\.", 2);
+		if(parts.length < 2)
+			filename += ".jpg";
```
mvdan: In any case, a bad guess is not worse than not guessing.

I would even argue it's better; It won't display properly, but at least it will show up in the gallery.

QuantumBadger:
Thanks @Bragado! I agree that this is better than the current situation -- I assume most image viewers will ignore the file extension and look at the actual data, so this should work fine in most cases.

If you want to improve this, the MIME type is available here:

https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java#L436

(line 436 -- `filenameFromString()` is called on line 438).

I'll merge this as-is for now though, because it's definitely an improvement! Thanks!

Issue #413, opened 2016-11-26
Share Images Directly


Nutomic:
When you click "share" on an image post, RedReader will only share the link to the image. If I want to share the image itself (eg in Whatsapp), I have to save the image, find it in a file browser, and then share it.

Instead, it would be great if the share action on image posts would directly share the image file.

fhtagn:
My simple proposal keeps the "Share" entry to mean sharing the link, and adds "Share Image" if the post is an image [as per](https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/common/LinkHandler.java#L303).

QuantumBadger:
Done by @fhtagn -- now in Alpha 97:

https://redreader.org/alpha/

PR #418, opened 2016-12-12
Issue #413 - share image directly

fhtagn:
Hey!

At a risk of contributing to [creeping featurism]( https://www.interaction-design.org/literature/book/the-glossary-of-human-computer-interaction/featuritis-or-creeping-featurism), I saw issue #413 and thought I could help. Partly because it seemed easy, and something I would actually use. I don't know (didn't look into it honestly) how to make a context menu entry off by default, but other than that this is working.

It will create an context menu entry to share the content of posts that have post.mIsProbablyAnImage == true

Please let me know if this needs some adaptation, if I missed something or otherwise feel free to re purpose the PR.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java
@@ -235,7 +235,7 @@ public final class RedditPreparedPost {
 		if(itemPref.contains(Action.GOTO_SUBREDDIT)) menu.add(new RPVMenuItem(activity, R.string.action_gotosubreddit, Action.GOTO_SUBREDDIT));
 		if(itemPref.contains(Action.SHARE)) menu.add(new RPVMenuItem(activity, R.string.action_share, Action.SHARE));
 		if(itemPref.contains(Action.SHARE_COMMENTS)) menu.add(new RPVMenuItem(activity, R.string.action_share_comments, Action.SHARE_COMMENTS));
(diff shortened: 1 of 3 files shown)


fhtagn:
This doesn't work with Whatsapp. Works for Hangouts, Telegram and Dropbox, though.

Maybe because the file is being shared with the .rr_cache_data extension from the cache?

Will get back to this later.

QuantumBadger:
Thanks, this looks good!

It would be an idea to only show the context item for images (similar to when saving images). See `RedditPreparedPost.java:234`:

    if(itemPref.contains(Action.SAVE_IMAGE) && post.mIsProbablyAnImage) ...

> I don't know (didn't look into it honestly) how to make a context menu entry off by default, but other than that this is working.

I'm happy for this to be on by default, so that's fine. Unfortunately, I think that if a user upgrades from an old version and has already modified this setting, it will be off by default for them (even if it's on by default for first-time users).

Maybe it would be worth checking if we're upgrading to version 77 (from version 76 or lower), and enabling the preference if so, similar to the following:

https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java#L188

Either way I'm happy to merge this in once the Whatsapp thing is cleared up, and the item is only shown for images. Thanks again :)

fhtagn:
Entry only shows for images, done.

As for Whatsapp, I would say Whatsapp is ignoring the mimetype, and using the extension to figure if it supports the file or not. So I resorted to saving the file (same as "Save image") and then share it.

Multiple shares of the same image will yield multiple copies in storage, which I don't like. But it seems it is either:

- this solution
- changing the cache so it stores the extension as the file extension
- ignoring Whatsapp

QuantumBadger:
No problem, that sounds like a reasonable solution.

I'll go ahead and merge this in -- thanks again!

Issue #267, opened 2016-01-02
Display size when loading images


QuantumBadger:
Under the circular progress bar, show the image size and progress, e.g. "1 MB / 15 MB".

QuantumBadger:
See https://github.com/QuantumBadger/RedReader/pull/269

QuantumBadger:
Done in Alpha 101.

PR #422, opened 2016-12-18
#267 Display size when loading images

Bragado:
Simple implementation of image size progress below the progress bar


src/main/java/com/github/lzyzsd/circleprogress/DonutProgress.java
```diff
@@ -84,6 +99,34 @@ public void setProgress(float progress) {
 		}
 	}
 
+	public void setProgress(long bytesRead, long totalBytes) {
+
+		if(this.internalState != Type.SIZE_PROGRESSBAR) {
+
+			mbStatus = new Paint();
```
QuantumBadger: Better to do this (and lines 107-112) in `initPainters()`.

src/main/java/com/github/lzyzsd/circleprogress/DonutProgress.java
```diff
@@ -84,6 +99,34 @@ public void setProgress(float progress) {
...
+			mbStatus.setColor(finishedStrokeColor);
+			mbStatus.setStyle(Paint.Style.FILL);
+			mbStatus.setStrokeWidth(unfinishedStrokeWidth);
+			mbStatus.setTextSize(40);
+			mbStatus.setTextAlign(Paint.Align.CENTER);
+			mbStatus.setAntiAlias(true);
+
+			this.internalState = Type.SIZE_PROGRESSBAR;
```
QuantumBadger: This extra state probably isn't needed

src/main/java/com/github/lzyzsd/circleprogress/DonutProgress.java
```diff
@@ -28,6 +31,16 @@
...
 
+
+	private enum Type { NO_SIZE_PROGRESSBAR, SIZE_PROGRESSBAR };
+
+	private Type internalState = Type.NO_SIZE_PROGRESSBAR;;
+
+	private long totalBytes;
+	private long bytesRead;
+
+	private final int B2MB = 1048576;
```
QuantumBadger: Could be static, also the name could be more descriptive (`bytesPerMegabyte`)

src/main/java/com/github/lzyzsd/circleprogress/DonutProgress.java
```diff
@@ -116,19 +159,40 @@ private int measure(int measureSpec){
 		return result;
 	}
 
+	public static float round(float d, int decimalPlace) {
+		BigDecimal bd = new BigDecimal(Float.toString(d));
```
QuantumBadger: Creating a new `BigDecimal` every time we receive some data is quite inefficient -- the conversion from bytes to MB could be done purely using integer arithmetic, e.g.:

		final int totalKilobytes = bytes / 1024;
	    final int totalMegabytes = totalKilobytes / 1024;
	    final int remainder = totalKilobytes % 1024;
	    final String result = String.format("%d.%02d MB", totalMegabytes, (remainder / 10));

The above should produce a result in the form "2.35 MB".

src/main/java/com/github/lzyzsd/circleprogress/DonutProgress.java
QuantumBadger: Never mind, I can see it is

src/main/java/com/github/lzyzsd/circleprogress/DonutProgress.java
```diff
@@ -84,6 +99,34 @@ public void setProgress(float progress) {
...
+
+			this.totalBytes = totalBytes;
+		}
+
+
+		this.bytesRead = bytesRead;
+
+		float progress = ((float)((1000 * bytesRead) / totalBytes)) / 1000;
+
+		if(Math.abs(progress - this.progress) > 0.0001) {
```
QuantumBadger: Good thinking, although we probably don't even need updates this frequently (e.g. `0.0001` could be changed to `0.01`).

src/main/java/com/github/lzyzsd/circleprogress/DonutProgress.java
```diff
@@ -84,6 +99,34 @@ public void setProgress(float progress) {
...
 
+	public void setProgress(long bytesRead, long totalBytes) {
+
+		if(this.internalState != Type.SIZE_PROGRESSBAR) {
+
+			mbStatus = new Paint();
+			mbStatus.setColor(finishedStrokeColor);
+			mbStatus.setStyle(Paint.Style.FILL);
+			mbStatus.setStrokeWidth(unfinishedStrokeWidth);
+			mbStatus.setTextSize(40);
```
QuantumBadger: This needs to be scaled for dp/sp. I probably would have just used a separate `TextView` rather than drawing the text directly, but this works too.

src/main/java/com/github/lzyzsd/circleprogress/DonutProgress.java
```diff
@@ -138,11 +202,25 @@ protected void onDraw(Canvas canvas) {
...
-			canvas.drawArc(finishedOuterRect, getStartingDegree(), getProgressAngle(), false, finishedPaint);
+
+			if(this.internalState == Type.SIZE_PROGRESSBAR) {
+
+				int width = getWidth() / 2 - (int) delta;
+				int height = getHeight() - (int) (delta * scale / 2 - delta);
+
+
+				canvas.drawArc(finishedOuterRect, getStartingDegree(), getProgressAngle(), false, finishedPaint);
+				canvas.drawText(round((float)bytesRead/B2MB, 1 ) + "MB / " + round((float)totalBytes/B2MB, 1) + " MB", width, height, mbStatus);
```
QuantumBadger: Inconsistent spacing before "MB"

QuantumBadger:
Thanks for the PR, and sorry for the delay! (things have been quite busy over Xmas)

I've added some comments to the diff -- if you could please get these resolved then I'll get this merged in :)

QuantumBadger:
Looks like this has been abandoned, and no longer merges cleanly :(

I've implemented this myself in a slightly different way here:

https://github.com/QuantumBadger/RedReader/commit/c7ec690cee5768506f082b63662df5327a2fee8e

Closing this PR.

PR #427, opened 2017-01-04
Fix preload before opening in external app

m3sv:
Should work for most links.


QuantumBadger:
Thanks! Looks like the pmd checks were failing -- it's possible to run it locally using `gradle pmd`.

m3sv:
Yeah, already found and fixed errors. Thanks.

PR #428, opened 2017-01-04
Fix preload before opening in external app

m3sv:
Should work with most links.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/ImageViewActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/ImageViewActivity.java
@@ -211,82 +211,7 @@ public class ImageViewActivity extends BaseActivity implements RedditPostView.Po
 					return;
 				}
 
-				CacheManager.getInstance(ImageViewActivity.this).makeRequest(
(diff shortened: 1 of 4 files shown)


QuantumBadger:
Thanks!

Are you sure this works for imgur though? I think the imgur API erroneously returns the `image/gif` mimetype for gifv files, which might be GIFs but are usually videos. We don't know which until actually fetching the image.

m3sv:
Indeed, it doesn't work. I'll look in to it.

QuantumBadger:
Maybe the `mp4` field returned by the imgur API could be of use. If it's set, then the mimetype should be `video/mp4`.

m3sv:
I already fixed the problem. Just need to commit.

QuantumBadger:
Excellent work, thank you. I'll merge this in now.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/48d38c83ddd839dac86ce4195201768d5a16468d

QuantumBadger:
Alpha 99 now up.

PR #447, opened 2017-03-01
Updated german l18n

Clubfan22:
Added a few strings for the german translation


QuantumBadger:
Danke schön :)

In future the new strings should go at the end of the file (they're grouped by date to make it easy to see which translations have been done). But I'm happy to accept this PR as-is, and I might reorganise it at some point. Thanks again!

PR #448, opened 2017-03-01
Added Context Menu for links

Clubfan22:
Long-pressing on a link button now shows a context menu with following actions:
* Copy URL to clipboard
* Share URL
* Open URL in external browser

In case a link is an image, two additional actions are available:
* Save Image
* Share Image

The actions are configurable through the menu preferences.

I moved some image-and-permission-related code from RedditPreparedPost into own classes to make it better reusable.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/image/SaveImageCallback.java
+++ b/src/main/java/org/quantumbadger/redreader/image/SaveImageCallback.java
@@ -1,0 +1,155 @@
+package org.quantumbadger.redreader.image;
+
+import android.content.Intent;
+import android.net.Uri;
+import android.os.Environment;
+import android.support.v7.app.AppCompatActivity;
+import org.quantumbadger.redreader.R;
+import org.quantumbadger.redreader.account.RedditAccount;
+import org.quantumbadger.redreader.account.RedditAccountManager;
+import org.quantumbadger.redreader.activities.BaseActivity;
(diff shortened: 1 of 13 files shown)


Clubfan22:
I don't know why the Travis build fails, everything builds fine on my windows machine.
It might be related to the Travis problems though: https://www.traviscistatus.com/

src/main/java/org/quantumbadger/redreader/image/SaveImageCallback.java
```diff
@@ -0,0 +1,155 @@
+package org.quantumbadger.redreader.image;
```
QuantumBadger: It looks like the new files should have a GPL notice at the top :)

QuantumBadger:
Thanks, this is excellent stuff! I've left one comment above that new files need a GPL notice (which can just be copied from one of the other files), but other than that this looks ready to merge in!

It looks like the German translation additions have been included in this PR as well, but that's fine. In future, keeping them in separate branches should avoid the conflict.

Clubfan22:
Sorry for missing the copyright headers, didn't notice it with IntelliJ's auto-collapse and thought it was set  up automatically, just added those. And yes, I accidentally started this branch off the the localizastion changes, won't happen next time :)

QuantumBadger:
Great, thanks! I'll merge this in now.

PR #451, opened 2017-03-02
Suggested subreddits for "Custom Location" now separated by account

### LINKED ISSUE #446: Make subreddit autocompletion account-specific
Subreddits visited/subscribed on one account shouldn't appear in the autocompletion list of another account.

https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/reddit/RedditSubredditHistory.java

Clubfan22:
Should fix #446

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java
@@ -290,7 +290,7 @@ public class MainActivity extends RefreshableActivity
 				break;
 
 			case MainMenuFragment.MENU_MENU_ACTION_PROFILE:
-				LinkHandler.onLinkClicked(this, new UserProfileURL(RedditAccountManager.getInstance(this).getDefaultAccount().username).toString());
+				LinkHandler.onLinkClicked(this, new UserProfileURL(username).toString());
 				break;
 
 			case MainMenuFragment.MENU_MENU_ACTION_CUSTOM: {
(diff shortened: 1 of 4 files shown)


QuantumBadger:
Thanks! Looks good, although I have a couple of minor comments:

I think `RedditAccount.getCanonicalUsername()` should be used instead of `RedditAccount.username`, to ensure that the lookup is case insensitive. Or even better, `RedditAccount` could probably be used as the key to the `HashMap` instead of `String` :)

It looks like the checks for `!SUBREDDITS.get(username).isEmpty()` should be inverted (at the moment it seems to be adding the defaults if the list is *not* empty!)

In fact, those checks are already done inside `putDefaultSubreddits()`, so no need to duplicate them:

	public static synchronized void addSubreddit(final RedditAccount account, final String name) throws RedditSubreddit.InvalidSubredditNameException
	{
		putDefaultSubreddits(account);
		SUBREDDITS.get(account).add(General.asciiLowercase(RedditSubreddit.stripRPrefix(name)));
	}

Generally fine though -- I'll merge once the above is fixed :)

QuantumBadger:
Actually it looks like there could be an infinite loop when calling `putDefaultSubreddits()` from `addSubreddit()`. Maybe better to replace line 45 `addSubreddit(username, subreddit)` with a call to `add(General.asciiLowercase(RedditSubreddit.stripRPrefix(name)))` on the `HashSet` directly.

Clubfan22:
Thanks for pointing out the flaws, should be fixed now!

QuantumBadger:
Thanks!

Travis is failing, but I think that's a Travis issue (failing to download a library). I'll go ahead and merge this in :)

PR #458, opened 2017-03-07
Added context menu to subreddits in mainmenu

Clubfan22:
This adds a context menu to the subreddit items in the main menu with following actions:

- Open in external browser

- Share

- Copy URL

- (if logged-in:) subscribe / unsubscribe

Added with b38fec5
- block / unblock

- pin / unpin

After unsubscribing from a subreddit, it might take a few seconds until the list is refreshed and that subreddit is removed. If the user is trying to unsubscribe from the same subreddit in the meantime, a small toast notification is shown.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/adapters/MainMenuListingManager.java
+++ b/src/main/java/org/quantumbadger/redreader/adapters/MainMenuListingManager.java
@@ -88,6 +88,11 @@ public class MainMenuListingManager {
 	public enum SubredditAction {
 		SHARE(R.string.action_share),
 		COPY_URL(R.string.action_copy_link),
+		BLOCK(R.string.block_subreddit),
+		UNBLOCK(R.string.unblock_subreddit),
+		PIN(R.string.pin_subreddit),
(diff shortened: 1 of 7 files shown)


QuantumBadger:
Great work as always, thanks!

This basically looks ready to merge, although it looks like there are a couple of commented-out uses of `setPinnedSubreddits()` and `setBlockedSubreddits()` -- are they supposed to be there?

It would be good to show a visual indication to the user that the subscription/etc is in progress, maybe using a toast, or even changing the item text in the main menu.

I'm happy to merge this as soon as you confirm it's ready. Thanks again :)

Clubfan22:
Those lines were just some left-overs from tests, I removed them now.
Also, added toast notifications for subscribing and unsubscribing, the other actions don't need these in my opinion as they take effect immediately.

QuantumBadger:
Excellent, thanks I'll merge this now.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/e52961c5e4868ca4affbafae0aa0400bde9aa749

PR #463, opened 2017-03-08
Update strings.xml (eo)

verdulo:


QuantumBadger:
Thanks, merging now.

When the language is set to Esperanto, does "flair" show up as "etikedo" on the reddit site, or do they officially leave it untranslated?

verdulo:
On https://www.reddit.com/r/Esperanto/ it's translated as 'etikedo' (label).

PR #468, opened 2017-03-23
Improved parsing of underscores

### LINKED ISSUE #465: Markdown parser doesn't handle underscores correctly
The markdown `A_B_C_D_E` is currently rendered A&zwnj;_B_&zwnj;C&zwnj;_D_&zwnj;E by the app, but reddit doesn't apply any formatting on the main site. The markdown parser should be changed to check whether or not the underscores have spaces next to them.

Clubfan22:
Should fix #465, but a few differences to Reddit's parser remain

I created a test post on reddit: https://www.reddit.com/r/test/comments/612w0s/testing_underscore_parsing/ , the app's parsing is now the same as Reddit's except the 7th and 8th example. Also, I don't think use cases for that are common.


Clubfan22:
5271cbe fixes #469 
From the commit message:

> Reddit markdown like `^(all of this is supposed to be in superscript)` is now being parsed correctly.
> 
> There still are differences between this implementation and Reddit's with parentheses within superscript:
> In cases like
> `^(this is superscript and contains ( opening and closing ) parentheses)`, Reddit's parser ends the superscript with the first occurence of a closing parenthese and displays the second one. This implementation considers balanced parentheses more importantly and displays the first closing parentheses within the superscripted part and uses the second one to end that part.
>
> This behaviour can be changed easily, but it remains questionable to me if Reddit's behaviour should be mimicked

An example can be found [here](https://www.reddit.com/r/test/comments/612w0s/testing_underscore_parsing/)

QuantumBadger:
Great work, thank you!

Unfortunately Travis says that one of the new tests is failing with an exception:

https://travis-ci.org/QuantumBadger/RedReader/builds/215017638#L1835

Once that's fixed, this looks ready to merge!

QuantumBadger:
Thanks -- I'll merge this in, although it looks like there aren't any tests for superscript text in parentheses (the existing tests are a bit lacking in general, so I can't really blame you :))

Clubfan22:
Writing tests for superscript in parentheses is a bit difficult - as far as I can see it - because it's not a change to the parser but to the MarkdownParagraph. It looks like one would have to mock at least the SpannableStringBuilder, at least that's what the test's exception says. Unfortunately, I don't really have any experience with that (yet).

QuantumBadger:
Okay, that's fine.

Changelog updated: https://github.com/QuantumBadger/RedReader/commit/bb5a220c938b50d59511e81383ffdb18032b3315

PR #471, opened 2017-03-26
Show caption in Imgur Album View as subtitle

Clubfan22:
This displays the image captions in the subtitle of images of imgur albums


QuantumBadger:
Thanks! I think this needs some visual improvement -- the caption text should be distinguished from the rest of the subtitle somehow (maybe in a different colour, and with a gap). Also the font colour isn't easily readable for so much text.

QuantumBadger:
This looks much better now, thanks! Merging this in.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/10f60aab0343a1b2900bee556d73d9ed40401578

Issue #472, opened 2017-03-29
Better visibility for Markdown formated code


Hund:
I think it's really hard to see code blocks [[Example](https://i.imgur.com/crplH7c.jpeg)]. I think they should have a light grey background or some other colour for better visibility.

PR #474, opened 2017-03-29
Code now has light-grey background color

Clubfan22:
See #472

Both inline code (\`) and code blocks (introduced with 4 spaces or a tab)
now have a light-grey background color for better distinction.

Ideally, it would be possible to set that color based on the currently selected
theme, but it looks like the code would have to be refactored for that.

For now, I've chosen \#33333333 because it seems to be working in all themes.

A test post can be found [here](https://www.reddit.com/r/test/comments/612w0s/testing_underscore_parsing/).

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/reddit/prepared/markdown/MarkdownParagraph.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/prepared/markdown/MarkdownParagraph.java
@@ -23,12 +23,7 @@ import android.support.v7.app.AppCompatActivity;
 import android.text.SpannableString;
 import android.text.SpannableStringBuilder;
 import android.text.Spanned;
-import android.text.style.ClickableSpan;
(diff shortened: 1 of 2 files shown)


QuantumBadger:
Thanks -- I haven't tested this yet, but presumably large blocks of four-space-indented code will look weird with this? I.e. the background will only extend to the end of the line, and so if different lines have different lengths then the background will end at inconsistent points?

Clubfan22:
The background actually extends to the end of the screen, it looks like this:
http://i.imgur.com/uv1FA4L.png

QuantumBadger:
That's only one line of text though -- I meant something like:

    a
    ab
    abc
    abcd

QuantumBadger:
Closing for now due to inactivity -- please re-open if this gets updated :)

Clubfan22:
Yes, I'll try to work on it again once I find the time, I'm a bit busy with university.

Issue #475, opened 2017-03-29
"Block subreddit" option in /r/all post context menu


QuantumBadger:
https://www.reddit.com/r/RedReader/comments/616lul/suggestion_for_blocking_subs/

QuantumBadger:
Added by @Clubfan22.

PR #477, opened 2017-03-30
Added block, pin and subscribe actions to RedditPreparedPost's context menu

Clubfan22:
Similiar to #458, I added menu items to (un)subscribe to a subreddit,
(un)block and (un)pin it, depending on its current state.

It's only activated when the subreddit's name is displayed.

This should help satisfy #475.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java
@@ -233,6 +233,27 @@ public class MainActivity extends RefreshableActivity
 					).apply();
 
 				}
+
+				if (lastVersion <= 77) {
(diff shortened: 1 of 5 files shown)


QuantumBadger:
Thanks! This generally looks good, but it's not clear that some of these menu items refer to the subreddit rather than the post. It would be good to rename them to "Subscribe to Subreddit" and "Pin Subreddit to Main Menu".

Also, all three items should appear directly after "Go to Subreddit" in the list.

QuantumBadger:
Thanks, merging in now.

QuantumBadger:
Changelog updated and some fixes added here: https://github.com/QuantumBadger/RedReader/commit/4f551ceab239299b7480ddd2b5ca0151a5335f88

PR #488, opened 2017-04-28
Bump gradle, pmd and some libs, clean up translations

mvdan:
I did not bump the support libs nor the target/compile SDK as those are more invasive, and they actually break the build (semver anyone?).


QuantumBadger:
Looks good to me, thank you!

PR #490, opened 2017-04-30
Replaced thumbnails with previews

md-mariok:
This change replaces the small thumbnails with large previews. Previews fill the entire column and are height capped to make sure the entire image for a post is visible without scrolling.

The preview images are not tap or fling aware, only the post title is to maintain the original feel of the app.

This is a big design change, I understand if you rather not change the app so dramatically. Another option is to allow users to select between thumbnail and previews. I personally really like the large previews, especially for subreddits like aww and pics.

This is how it looks on the front page

screen shot 2017-04-29 at 5 14 24 pm

Even subreddits like gadgets look great using previews

screen shot 2017-04-29 at 5 15 07 pm

If no images are allowed the subreddit looks just like it does today

screen shot 2017-04-29 at 5 18 56 pm

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,7 @@ dependencies {
     compile 'com.android.support:design:25.0.1'
     compile 'com.squareup.okhttp3:okhttp:3.6.0'
     compile 'info.guardianproject.netcipher:netcipher:1.2.1'
+	compile 'com.android.support.constraint:constraint-layout:+'
 
     testCompile 'junit:junit:4.12'
 }
--- a/src/main/java/org/quantumbadger/redreader/jsonwrap/JsonBufferedObject.java
+++ b/src/main/java/org/quantumbadger/redreader/jsonwrap/JsonBufferedObject.java
@@ -17,9 +17,13 @@
 
 package org.quantumbadger.redreader.jsonwrap;
 
+import android.util.Log;
+
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
+
+import org.quantumbadger.redreader.reddit.things.RedditPost;
 
 import java.io.IOException;
 import java.lang.reflect.Field;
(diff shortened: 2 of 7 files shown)


mvdan:
I very much prefer thumbnails. The version with no pictures is okay-ish, but I prefer it with the thumbnails.

I would say make this configurable. Whether or not the default is kept at thumbnails or switched to previews is up to @QuantumBadger.

QuantumBadger:
Thank you -- I like how this looks a lot, but I agree with @mvdan that there should be a preference for this, and I think the default should be the existing layout.

Does this keep all the preview images in RAM after downloading them? It's okay to do this with thumbnails because they're much smaller, but since the previews are larger we could run into memory issues with lists of 100+ posts.

Also, how does this look in landscape mode? If the preview aspect ratio is preserved then each post will presumably be bigger than the screen :)

Generally this is a good idea, and if the above points are fixed I'd be happy to merge this in.

md-mariok:
I can make it an option and leave the default as is. I will let you decide how to allow users to switch between layouts.

As for the memory question, the RecyclerView should take care of reusing views and only load the images that are going onto the screen. I haven't read the RecyclerView adapter logic though so I am not sure how's currently implemented. It does use more storage and bandwidth to download previews.

Landscape mode works as expected, that's the job of the PreviewView. Differently from thumbnails, previews are not resized on download. They are stored in raw format and dynamically resized on display by the PreviewView.

QuantumBadger:
Thank you :)

While I remember, it looks like Gradle is being told to use the latest version of `com.android.support.constraint:constraint-layout`. Unfortunately Android support library changes often break things, so this should be tied to a specific version (the same one as the other support libs).

> the RecyclerView should take care of reusing views and only load the images that are going onto the screen

It seems to keep a reference to every bitmap in `RedditPreparedPost.thumbnailCache`, although maybe I'm mistaken.

md-mariok:
The bitmaps are indeed kept in memory by RedditPreparedPost.thumbnailCache. I will submit a separate patch for that first.

ikocevski:
Any news on this?

QuantumBadger:

rakshazi:
What a pity, that's really great function :(

PR #497, opened 2017-05-12
Reducing Listing Clutter

neilweber:
On my 4.5" display, I found the listing to be cluttered.  To address this, I removed the userid that submitted the link and changed the post time to display only one unit instead of two (e.g. "10 mins, 30 secs ago" is now just "10 mins ago.")


QuantumBadger:
Thank you for the pull request! Because this would be quite a significant change to the app's appearance (which not all users would be happy with), there would need to be a preference for this before I can merge it in.

neilweber:
Are you thinking a separate preference for each change or one for both?

QuantumBadger:
> Are you thinking a separate preference for each change or one for both?

Separate preferences for these would be great. Ideally there would be a list preference similar to the one in `Settings > Appearance > Comment Header`, allowing the user to completely customise the post subtitle, although for now a simple checkbox would be fine.

QuantumBadger:
Here's a quick guide for adding a preference:

https://github.com/QuantumBadger/RedReader/wiki/%5BDev%5D-Adding-a-Preference

PR #502, opened 2017-05-21
No longer caching thumbnails in memory

md-mariok:
The RecyclerView takes care of managing posts on the screen. Whenever a
post is shown its thumbnail is loaded from the file cache. Since that's
done asynchronously there is no performace hit.

Images in this patch are resized by the ImageView. This is not stricly
necesary but it makes the display logic simpler and would allow for
the size of thumbnails to be configurable.

This patch reduces memory utilization which is needed by the patch implementing preview images

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/image/ImageLoader.java
+++ b/src/main/java/org/quantumbadger/redreader/image/ImageLoader.java
@@ -1,0 +1,74 @@
+/*******************************************************************************
+ * This file is part of RedReader.
+ *
+ * RedReader is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * RedReader is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with RedReader.  If not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************/
+
+package org.quantumbadger.redreader.image;
+
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.util.Log;
+import android.util.Pair;
+import android.widget.ImageView;
+
+import org.quantumbadger.redreader.cache.CacheManager;
+import org.quantumbadger.redreader.common.AndroidApi;
+
+import java.util.concurrent.LinkedBlockingDeque;
+
+/**
+ * Created by Mario Kosmiskas on 5/21/17.
+ */
+
+public class ImageLoader {
+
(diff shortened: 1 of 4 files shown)


QuantumBadger:
Thanks for this -- it looks like Travis is failing due to pmd.

> Since that's done asynchronously there is no performace hit.

I've taken a look through the code and it seems that the IO is done on the UI thread, in the following two places:

* In the constructor of `RedditPostView`, the `thumbnailHandler` is created using the UI thread looper (`Looper.getMainLooper()`), meaning that `BitmapFactory.decodeStream` is called from the UI thread
* In `RedditPostView.reset()`, a runnable is posted to the UI thread (line 282) which calls `getThumbnail()`, which in turn calls `BitmapFactory.decodeStream`.

Doing IO (and/or image resizing) on the UI thread could have a serious effect on performance and so unfortunately I can't accept this PR as-is.

If this were changed to asynchronously load thumbnails, then it would need to load a few in advance (from outside the top and bottom of the visible area) to ensure that they don't "pop in" when the user scrolls. So it wouldn't be able to rely purely on the `RecyclerView` requesting the view.

I'd be happy to accept a modified version of this which fixed the above issues (e.g. by having a single new thread with its own Looper, responsible for loading the thumbnail images from disk, and having the post listing decide which thumbnails to load based on `RecyclerView` scroll position). Please let me know if you have any questions :)

md-mariok:
Sure, moved I/O into a background thread

src/main/java/org/quantumbadger/redreader/image/ImageLoader.java
```diff
@@ -0,0 +1,74 @@
...
+		public void run() {
+			while (true) {
+				try {
+					final Pair task = mQueue.take();
+					final Bitmap image = BitmapFactory.decodeStream(((CacheManager.ReadableCacheFile)task.first).getInputStream());
+
+					AndroidApi.UI_THREAD_HANDLER.post(new Runnable() {
+						@Override
+						public void run() {
+							((ImageView)task.second).setImageBitmap(image);
```
QuantumBadger: If the `ImageView` is re-used for another post thumbnail in the meantime, then presumably this will encounter a race condition (where the wrong thumbnail could temporarily be shown for a post)? I.e.:

* `loadImage(imageA, imageView1)` is called
* Before the loading is complete, imageView1 is re-used by the RecyclerView for another post, and so `loadImage(imageB, imageView1)` is called
* `loadImage(imageA, imageView1)` completes, and imageView1 is set to imageA
* `loadImage(imageB, imageView1)` completes, and imageView1 is set to imageB

So the thumbnail will briefly flicker to show an incorrect image.

QuantumBadger:
Thanks, I've left a code comment about what looks like a race condition.

I don't think this change fixes the following issue though:

> If this were changed to asynchronously load thumbnails, then it would need to load a few in advance (from outside the top and bottom of the visible area) to ensure that they don't "pop in" when the user scrolls. So it wouldn't be able to rely purely on the RecyclerView requesting the view.

The images are only loaded off disk after the post has already appeared on screen. So the thumbnails will "pop in", and I'm guessing the post view is resized to accommodate it?

QuantumBadger:

PR #503, opened 2017-05-23
Added setting for self-post collapse

Clubfan22:
It was requested on reddit: https://www.reddit.com/r/RedReader/comments/6bv9ma/an_option_to_turn_off_collapsing_of_selfposts_on/

For now, one can choose between collapsing the text or doing nothing, but it can be (almost) trivially extended to support more actions.

One might also want to refactor the self-post code in the future to be treated the same as a comment in some cases.


QuantumBadger:
Great, thank you! Merging in now.

> One might also want to refactor the self-post code in the future to be treated the same as a comment in some cases.

Possibly, although there are a few differences (e.g. long-click menu actions, swipe actions, etc).

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/0915f27fbdc752740d1a47771b3dcc22f37b793d

PR #509, opened 2017-05-27
Master

awells111:
I moved UI_THREAD_HANDLER out of AndroidApi.java and deleted the whole class along with ImageTileSourceBitmapRegionDecoder which was unused and targeting API 10.  I also removed many @SuppressLint("NewApi") and @TargetApi calls that were targeting API levels 14 and below.


QuantumBadger:
Thanks, merging this in!

I'll move the handler into `General` instead of `Constants`, but other than that this is fine.

QuantumBadger:
https://github.com/QuantumBadger/RedReader/commit/a814ef7313851c461c58c7904595545af4c1d679

Updated changelog here: https://github.com/QuantumBadger/RedReader/commit/d20dc68dfe5467e2af2fa60979f0244422462cf9

Issue #387, opened 2016-09-11
Add controls for videos/GIFs


ShadowNinja:
This is useful particularly with long videos, where you may want to skip around or re-watch a short segment without wating for the entire thing to repeat (or reopening and starting from the beginning).

polyzen:
I think this can be closed.

QuantumBadger:
Yep, this has been implemented

PR #511, opened 2017-06-23
Add video playback controls

ahomphophone:
Implemented #387. I added a media controller to the video view. Looks like all the playback logic was already in MediaVideoView. I just needed to create and set the media controller. 

This also disables the ability to finish the activity when tapping the video because that is how the playback controls appear and hide. I'm not sure if this should be a preference but my instinct is to tap on the video to bring up playback controls and just pressing the back button to exit.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/ImageViewActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/ImageViewActivity.java
@@ -424,7 +424,10 @@ public class ImageViewActivity extends BaseActivity implements RedditPostView.Po
 							final MediaVideoView videoView = new MediaVideoView(ImageViewActivity.this);
 
 							videoView.setVideoURI(cacheFile.getUri());
-							videoView.setMediaController(null);
+
+							MediaController mediaController = new MediaController(ImageViewActivity.this);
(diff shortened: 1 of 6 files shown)


QuantumBadger:
Thanks!

Before I merge this in, please could you make it a preference (off by default) so that people can keep the old behaviour if they want?

ahomphophone:
Ok. Added preference.

QuantumBadger:
Great, thank you!

Sorry to be a pain, but it looks like there's one more thing: the changes to `BasicGestureHandler` break the abstraction (by generating two events for the same tap).

It should be possible to move the `performClick()` call (which us currently inside `BasicGestureHandler`) to the `onSingleTap()` callback in `ImageViewActivity`, if you make the `videoView` a member. No changes should be necessary to `BasicGestureHandler`.

I promise I'll merge it in once that's fixed :stuck_out_tongue:

ahomphophone:
Ok sorry! Done.

QuantumBadger:
Thank you! Merging in now.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/ded5c57aab5f1a6f62d2b3a40f25788e9ae4a0eb

Issue #520, opened 2017-07-24
Add support for v.reddit.com


Clubfan22:
[Per Reddit](https://www.reddit.com/r/RedReader/comments/6p4agf/vreddit_links_not_working/)

From what I've seen so far, *v.reddit.com* is Reddit's video host. When clicking on a post linking there. the website embeds the video into the OP. Currently, RedReader just opens the thread without showing the video.

An example for this behaviour can be found at: https://v.redd.it/smasrs6l7ebz

Technically, the video appears to be based on [https://en.wikipedia.org/wiki/Dynamic_Adaptive_Streaming_over_HTTP](DASH).

For RedReader, I think having an extra activity for reddit videos would be the way, similar to GIFs and imgur albums.

QuantumBadger:
Fixed in Alpha 137 :)

PR #521, opened 2017-07-24
#520: v.reddit.com links are opened externally by default

Clubfan22:
Until a better solution is in place, I think opening v.reddit.com links
externally is the least problematic way.


QuantumBadger:
Thanks for the pull request! I've just put in a fix for this here, which uses the internal video player:

https://github.com/QuantumBadger/RedReader/commit/c4e39ff66e2c8f178fc4a6304c43f102c1e863a5

PR #543, opened 2017-11-01
Added checkbox for disabling "replies to inbox" when replying to comm…

### LINKED ISSUE #286: Make it possible to disable inbox replies when commenting

Clubfan22:
…ents

When replying to comments, it's now possible to disable that replies will be sent to the inbox. (Potential improvement: preference for default behaviour)

Reddit's API endpoint /sendreplies is used for this.
About the actual implementation, I'm not sure if it would be more elegant to move the actual call into CommentReplyActivity, but then we have to escalate the id of the newly created post, which would entail some refactoring.

On another note, Reddit's API responses surely have to be a joke, right? No sane person could do that...

Fixes #286

### DIFF
--- a/assets/changelog.txt
+++ b/assets/changelog.txt
@@ -2,6 +2,7 @@
 Fix for fullscreen video in internal browser (thanks to Clubfan22)
 Preference to exclude text when sharing comment (thanks to Clubfan22)
 Scroll to top when collapsing self post (thanks to ajgoda90)
+Disable sending replies to inbox for comments (thanks to Clubfan22)
 
 82/1.9.8.3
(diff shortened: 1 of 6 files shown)


QuantumBadger:
Looks good to me, thanks!

> On another note, Reddit's API responses surely have to be a joke, right? No sane person could do that...

I assume you're referring to this:

	return "t1_" + response.asObject().getArray("jquery").getArray(30)
			.getArray(3).getArray(0).getObject(0)
			.getObject("data").getString("id");

That does look bad, even by reddit's standards :)

Are those array indexes always the same? (`getArray(30)`, etc) "30" seems like a weird choice of number :P

PR #549, opened 2017-11-27
Added support for /r/random

### LINKED ISSUE #200: Issues with /r/random
I noticed that entering random into the custom subreddit bar produces pecular behavior.  It appears that the app polls /r/random once for the content and once for the title banner.  The result is that the banner and the content wind up coming from two entirely different subreddits.

AwIDoK:
Fixes #200


QuantumBadger:
Excellent, thank you! I'll merge this in.

New strings should go at the bottom of strings.xml (to assist translators), but I'll move them myself. Also be careful with tabs vs spaces :)

PR #574, opened 2018-02-03
Add search option in the "Custom Location" dialog

samdroid-apps:
To search Reddit, users currently must tap a subreddit (eg. /r/all), tap
the menu icon then select search.

This adds search an option in the "Custom Location" dialog; which
opens the search page with the query.  It searches all subreddits.


QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/b2bae6f53f7619c83f6d3c273e9001f681472f45

PR #576, opened 2018-02-07
Support large (18x9) aspect ratios.

ktechmidas:
Hi! Great work, I've been using this app for years. Unfortunately I've noticed an issue, I got a Mi Mix 2 today which is a bit unusual in that it's 18x9 instead of the regular 16x9 and also has an immersive mode. Most apps account for this and take up the full screen size available.

Unfortunately RedReader doesn't, while the top image looks okay, when the nav bar disappears it leaves an empty black space (as you can see in the next one)

[I found a blog post](https://android-developers.googleblog.com/2017/03/update-your-app-to-take-advantage-of.html) from Google which tells us how to support a larger display format. I added the line from the post into the manifest and rebuilt and it solved the aspect issue for the Mi Mix 2 (and probably other 18x9 phones). - You can see below for the results.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -40,6 +40,7 @@
                  android:largeHeap="true"
                  android:hardwareAccelerated="true"
                  android:allowBackup="true">
+		<meta-data android:name="android.max_aspect" android:value="2.1" />
 
 		<meta-data android:name="com.lge.support.SPLIT_WINDOW" android:value="true" />
 


philipwhiuk:
See https://stackoverflow.com/questions/43179314/where-does-android-max-aspect-meta-make-difference

ktechmidas:
@philipwhiuk we're using SDK 23, so it will have an effect here. The alternatives are many though...

> NOTE: You do not need to set a maximum aspect ratio if an activity's android:resizeableActivity attribute is set to true. If your app targets API level 24 or higher, this attribute defaults to true.

So we have three choices to fix this.. 
1) We can do what I did in the pull request.
2) We can bump the SDK level to 24
3) We can set android:resizeableActivity to true

QuantumBadger:
Thanks, looks good to me! Merging in.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/3f05b4dbc513eb5519bece7d1d0e091e8d9311c3

PR #578, opened 2018-02-11
Improvements to the sharing system.

c0d3st0rm:
As the message says, this commit adds a new "Sharing" category under "Behaviour" in settings, moves two of the sharing options to the new category ("Share as permalink" and "Include text when sharing comment"), and introduces a new "Include title/description when sharing" option, which allows easy link-only sharing by optionally (instead of the default, as it was previously) including the "Comments for <post name>" etc text when shared. Also localized the hardcoded "Comments for <post name>" and "Comment by <author> on Reddit" strings.


QuantumBadger:
Nice work, thanks! I'll merge this in.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/249a2d20bceb8fbdfa2af803b834a26d9f0917b0

PR #590, opened 2018-04-12
Update Dutch translations

Ghostkeeper:
I got myself a new app on my phone, Stringlate. It finds repositories that need their localisation texts updated. So I did that on the way home from work today. Quite fun!

It generated this file. Maybe this is something you'd like to add?


QuantumBadger:
Thanks! Looks good to me -- I'll merge this in.

PR #591, opened 2018-04-22
Fix Search Bug and Update Libraries

ianmcxa:
I noticed that searches would not include nsfw results, even if `show nsfw content` was enabled. This seems to be due to an undocumented API change requiring a cookie for searches to return nsfw results. I've implemented this with the okhttp backend. I have also updated the SDK version to 27.


QuantumBadger:
Great, thank you!

Any chance this could be tied to the existing NSFW preference please?

https://github.com/QuantumBadger/RedReader/blob/master/src/main/res/xml/prefs_behaviour.xml#L158

Some people may have NSFW enabled on their account, but disabled in the app (because you don't necessarily want it enabled on the bus :)) Also I assume this works if someone is logged out?

It looks like the build is failing due to the gradle wrapper change -- please could you fix the `distributionSha256Sum` in `gradle-wrapper.properties`? Just downloaded it and got the checksum `7a2c66d1a78f811d5f37d14630ad21cec5e77a2a4dc61e787e2257a6341016ce`.

ianmcxa:
The NSFW preference will prevent NSFW results showing up, even with this enabled, so we shouldn't need to tie this to the NSFW preference.

ianmcxa:
Looks like the build is still failing. It's saying licenses have not been accepted. I'm not sure how to fix that.

QuantumBadger:
I think the `.travis.yml` file needs updating:

https://github.com/QuantumBadger/RedReader/blob/master/.travis.yml

It tells Travis which SDKs to download, and it's still stuck on the old versions.

ianmcxa:
I updated .travis and it looks like it's still failing.

QuantumBadger:
Travis still seems to be downloading the wrong SDK -- maybe it checked out the wrong commit somehow. Restarting the build didn't help.

Would you mind adding another commit to the pull request please? Just something minor like an extra newline somewhere, to hopefully make Travis detect the latest commit? If that doesn't work I'll probably just merge this and fix any actual issues which crop up.

ianmcxa:
Alright, pushed

QuantumBadger:
Perfect, thanks, looks like it's working now :) I'll get this merged in.

QuantumBadger:
Changelog updated:

https://github.com/QuantumBadger/RedReader/commit/e42ac202cde0ecf731a7381ac59b5c05b2b5ffcc

QuantumBadger:
Alpha 156 built.

PR #603, opened 2018-06-23
Add a blocklist to filter posts by (parts of) post.url

MoriTanosuke:
This allows users to hide post based on their URLs, to filter out unwanted content sources. The blocklist can be edited via preferences.

During my usage of reddit I often want to filter out memes, for example. And those tend to come from certain hoster URLs and I just want to get rid of all of them. With this change I can add (for example) `"gfycat.com"` to the blocklist and get rid of all of them at once.

This is my first attempt at a new feature for *RedReader*. I tried to keep the change as small as possible and follow the general layout of the code. If there is anything wrong with the code or someone can give me tips how to improve, I appreciate all comments.

I only do Android development occasionally, so I couldn't really figure out how to have a nice preferences screen, where the user can add/remove entries to the blocklist, so I added the preferences value as a simple `EditTextPreference` where I can at least enter the blocklist entries separated with `;`... I'd like to improve this, maybe someone can give me a link to the Android docs or into existing *RedReader* code.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -133,6 +133,13 @@
         <activity android:name=".activities.InboxListingActivity"
                   android:configChanges="orientation|screenSize|keyboardHidden"/>
 
+		<activity android:name=".activities.BlocklistListingActivity">
+			<intent-filter>
+				<action android:name="org.quantumbadger.redreader.activities.BlocklistListingActivity" />
+				<category android:name="android.intent.category.DEFAULT" />
+			</intent-filter>
+		</activity>
+
         <receiver android:name=".receivers.NewMessageChecker"/>
         <receiver android:name=".receivers.RegularCachePruner"/>
         <receiver android:name=".receivers.BootReceiver">
--- a/src/main/java/org/quantumbadger/redreader/activities/BlocklistListingActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/BlocklistListingActivity.java
@@ -20,19 +20,18 @@ public class BlocklistListingActivity extends BaseActivity {
 
 	private List<String> blockedPostUrls;
 
-	protected ListAdapter mAdapter;
-	protected ListView mList;
+	private ListAdapter mAdapter;
+	private ListView mList;
 
-	private Handler mHandler = new Handler();
 	private boolean mFinishedStart = false;
 
-	private Runnable mRequestFocus = new Runnable() {
+	private final Handler mHandler = new Handler();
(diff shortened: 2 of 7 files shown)


QuantumBadger:
Good idea, thanks!

One thing which would be worth improving is to filter the domain rather than the string as a whole.

For example, if you add `imgur.com` to the list, you want to filter:

```
http://imgur.com/abcdef
http://i.imgur.com/abcdef.jpg
```

But you don't want to filter:

```
http://imgur.com.example.com/
http://example.com/example_imgur.com
```

See `Uri.parse(string).getHost()` to find the domain, and maybe check using:

```
host.endsWith(".imgur.com") || host.equalsIgnoreCase("imgur.com")
```

But be careful to catch errors if the parse fails :)

> I couldn't really figure out how to have a nice preferences screen, where the user can add/remove entries to the blocklist

Same here, I don't think Android provides a way of doing this easily. If you're feeling ambitious, maybe have the preference launch a new Activity, containing a list of the existing entries, and an "add" button in the action bar.

See here for an example of launching an activity from a preference:

https://github.com/QuantumBadger/RedReader/blob/436cef98716ffad603dee0e04fe29516ee71836a/src/main/java/org/quantumbadger/redreader/settings/SettingsFragment.java#L189

Thanks again, and let me know if you have any questions.

MoriTanosuke:
I thought about parsing the URL for the host, I'll add that now.

And I think I'll give a shot at the preference activity to maintain the blocklist afterwards. :smile: Thanks for your feedback!

MoriTanosuke:
I rebased my fork with the current master and pushed a new commit replacing the simple text input with a slightly better `ListActivity` which displays the entries of the blocklist as a vanilla android `ListView`. The user can now add entries via a `Button` and remove entries with a single click on an existing entry.

device-2018-08-04-232212

I'm not 100% happy, because the plain `Button` at the top is still ugly. I'll have to check how *RedReader* handles this type of button in other places. If anyone has tips how to improve this, I'd be more than happy to replace this with something more suitable.

QuantumBadger:
I was about to suggest the `FloatingActionButton`, but it looks like you've beaten me to it :) An alternative would be to have an item in the list called "Add new item...", with a "plus" icon next to it, but the floating button is fine too.

Based on that screenshot, the activity would probably benefit from an action bar (with title) at the top, and maybe a message saying "no items" when the list is empty. Looking like good progress though!

MoriTanosuke:
I think I have to abandon this PR. 🤷‍♂️ Feel free to close the PR. I'll leave my branch up in case anyone discovers this and wants to pick it up again.

QuantumBadger:

PR #611, opened 2018-09-10
Added DeviantArt support

sharkwouter:
Here is the DeviantArt support I told you I wanted to work on on Reddit. Since the DeviantArt api doesn't give the description of images, I use the tags instead. The API also doesn't return the size, so I just return 0 for that. Hope this helps.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/LinkHandler.java
+++ b/src/main/java/org/quantumbadger/redreader/common/LinkHandler.java
@@ -440,7 +440,8 @@ public class LinkHandler {
 			reddituploadsPattern = Pattern.compile(".*[^A-Za-z]i\\.reddituploads\\.com/(\\w+).*"),
 			redditVideosPattern = Pattern.compile(".*[^A-Za-z]v.redd.it/(\\w+).*"),
 			imgflipPattern = Pattern.compile(".*[^A-Za-z]imgflip\\.com/i/(\\w+).*"),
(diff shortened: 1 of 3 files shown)


sharkwouter:
In case you are interested, the API documentation can be found here: https://www.deviantart.com/developers/oembed

It seems to be outdated, though. I only implemented the url type I could find and test.

QuantumBadger:
Thank you, this looks good to me! Sorry I didn't see your message on reddit, I assume you solved the build issue :)

The only concern I have is the regex:

```
deviantartPattern = Pattern.compile("https://www\\.deviantart\\.com/([\\w\\-]+)/art/([\\w\\-]+)"); 
```

This will only match the `www` subdomain (e.g. it won't match `https://deviantart.com`), and only `https` (as opposed to `http`). It also doesn't ignore characters at the end (e.g. if the URL ends in a slash).

So maybe worth making the regex more permissive to match the others, e.g.:

```
deviantartPattern = Pattern.compile(".*[^A-Za-z]deviantart\\.com/([\\w\\-]+)/art/([\\w\\-]+).*"); 
```

Having said that, I've just tested some links in a web browser, and the non-standard URLs seem to redirect to the "standard" one. So I'm happy to merge this as-is. Thanks again!

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/eb1d4d6e2a4abc2dbd74dfdf242054b2d5383f6d

Issue #558, opened 2017-12-31
Suggestion: Hide visited posts


p3lim:
Kind of like how RES does this, where you can hide all posts that you've visited, or only show those you've already visited.

asoback:
Is this what you were looking for? Posts are still visible after they have been clicked, until the next refresh.

p3lim:
When a link/post is visited it should not show up the next time the subreddit is refreshed/reloaded.
If that's what your commit does, then yes.

QuantumBadger:
Closing as fixed, thank you @asoback :)

asoback:
Thank you! This is my first true contribution to an open source project! I look forward to making many more!

PR #612, opened 2018-09-10
Hide visited posts option

asoback:
Added behavior preference to not load/display visited posts.

In reference to issue #558

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
 		google()
 	}
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.1'
+        classpath 'com.android.tools.build:gradle:3.1.4'
     }
 }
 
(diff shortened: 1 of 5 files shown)


QuantumBadger:
Thank you, this looks good to me!

QuantumBadger:
I've inverted the preference to make it clearer -- it's now "hide read" rather than "show read": https://github.com/QuantumBadger/RedReader/commit/2bf7cddbb5b11c5be1e21bff73e8c95a58c5a667

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/d22fad36e9901c8b8227ce1f11618324f12fab9a

Ty2357364:
This is awesome. However, it works for a while, but then it forgets which ones I've read. (I like to view top of all time; am I the only one? The best posts come back and I have to click thru them again.)
Is there a limit on the number of posts that can be remembered as read? I'm sure there is a limit on the number of read posts you can store; could you store an index to jump to instead, like, "on this subreddit, the top 400 posts of all time have been read."? I'd miss new top posts, but it would be a nice option.

QuantumBadger:
@Ty2357364 Thanks! There's a setting for how long that info is kept under "Cache > Delete cached flags after".

Ty2357364:
@QuantumBadger Thank you! (Now I just wish I could keep the data forever and import it if I get a new phone... But I don't think that's realistic.)

PR #614, opened 2018-09-20
Added RedditAccount tests

JonaszJestem:


philipwhiuk:
Code-style doesn't match the rest of the project.

src/test/java/org/quantumbadger/redreader/account/AccountUtils.java
```diff
@@ -0,0 +1,23 @@
+package org.quantumbadger.redreader.account;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class AccountUtils
+{
+	static RedditAccount createAccount( String username )
```
philipwhiuk: Using the utils provides very little benefit and obscures the fact that there are two other arguments that you could be testing.

Issue #618, opened 2018-10-05
Make all main menu items hideable


QuantumBadger:
Currently it's possible to hide /r/random and /r/popular from the main menu. I've had a request to be able to hide the other items as well (/r/all, custom location, etc).

LizardWithHat:
I'd like to do this. Should the "subscribed subreddits" and multireddit list be hideable aswell?

QuantumBadger:
Thanks! Yes, feel free to make those hideable as well. It might be worth moving all the main-menu preferences to a sub-menu, to avoid the preferences getting too cluttered.

Let me know if you have any questions.

Cguy7777:
@QuantumBadger Looks like this has been added. :-)

PR #619, opened 2018-10-06
Added hideable Main Menu items

LizardWithHat:
This addresses #618.
I put the first five menu items in a sub-menu to not overload the menu settings screen too much.
This is my first PR, I'd appreciate any helpful tips and criticism.

screenshot_20181006-195632
screenshot_20181006-195622

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -1089,4 +1089,20 @@
 	<string name="pref_behaviour_hide_read_posts_key" translatable="false">pref_behaviour_hide_read_posts</string>
 	<string name="pref_behaviour_hide_read_posts_title">Hide read posts</string>
 
+	<!-- 2018-10-06 -->
+	<string name="mainmenu_shortcut_all">All subreddits</string>
+	<string name="mainmenu_shortcut_custom">Custom Location</string>
+	<string name="mainmenu_shortcut_frontpage">Frontpage</string>
+	<string name="mainmenu_shortcut_random">Random</string>
+	<string name="mainmenu_shortcut_popular">Popular</string>
+	<string name="pref_menus_mainmenu_shortcutitems_title">Main Menu Shortcuts</string>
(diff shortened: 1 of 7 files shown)


src/main/res/values/strings.xml
```diff
@@ -1089,4 +1089,20 @@
 	<string name="pref_behaviour_hide_read_posts_key" translatable="false">pref_behaviour_hide_read_posts</string>
 	<string name="pref_behaviour_hide_read_posts_title">Hide read posts</string>
 
+	<!-- 2018-10-06 -->
+	<string name="mainmenu_shortcut_all">All subreddits</string>
+	<string name="mainmenu_shortcut_custom">Custom Location</string>
+	<string name="mainmenu_shortcut_frontpage">Frontpage</string>
+	<string name="mainmenu_shortcut_random">Random</string>
+	<string name="mainmenu_shortcut_popular">Popular</string>
```
QuantumBadger: For the five strings above, I think we should just use the existing ones (`mainmenu_frontpage`, `mainmenu_all`, etc)

QuantumBadger:
Thanks, this generally looks good to me. I've added one minor comment.

My only concern about this is that, because the existing preferences are being replaced with the `MultiSelectListPreference`, the user's existing preferences for this will be discarded. For example, if the user had decided to show `/r/popular` and `/r/random`, these will now be hidden after the update.

One way to solve this would be to check the old value of the preferences when starting up after an update, and adjust the new preference accordingly. See:

https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/activities/MainActivity.java#L231

Other than that, this looks great -- thanks for contributing :)

LizardWithHat:
Thanks for your insight.
I'm not sure if re-adding the two methods to get the old preferences was the best way, but it seemed like the most fitting one.

QuantumBadger:
Looks great, thanks! Merging this in.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/d7efbba06ff072b6d188f994bf2d85ba71e43cc1

PR #628, opened 2018-10-26
Update SDK version to 28

ianmcxa:
Just keeping up to date with the SDK

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
 		google()
 	}
     dependencies {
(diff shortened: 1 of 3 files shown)


QuantumBadger:
Thanks. I'd be a bit nervous about merging this in until the app has been thoroughly tested on Android 9:

https://developer.android.com/about/versions/pie/android-9.0-changes-28

Also the Travis script will need updating to fix the build:

https://github.com/QuantumBadger/RedReader/blob/master/.travis.yml

PR #640, opened 2018-12-11
No-wrap gold gilding badge

marcotc:
I noticed today on my phone that the `Gold xN` gilding badge can wrap between the strings `Gold` and `xN`.
Here's an example (I inserted artificial white-space to force it to happen in the emulator):
before

This PR replaces the regular space character between those two strings with a non-breakable space, with this as a result:
after


QuantumBadger:
Great, thanks! Merging this in.

PR #648, opened 2019-01-27
Added an indicator for the aspect ratio of media being loaded.

### LINKED ISSUE #632: [Feature Request] Show an indicator of orientation for loading media
It would be nice if the loading screen for pictures, GIFs, and videos indicated the orientation of what's loading (portrait/landscape/square), so that we can rotate our devices as desired before it's ready to view.

Cguy7777:
I decided to take a stab at actually implementing this idea, and was partially able to! The only issue is that as written it only works if the image/GIF/video being loaded has its height and width provided by an API, but that still manages to include a great deal of posts with Imgur, Gifycat, DeviantArt, and Streamable being supported. Partially closes #632.

### DIFF
--- a/src/main/java/com/github/lzyzsd/circleprogress/DonutProgress.java
+++ b/src/main/java/com/github/lzyzsd/circleprogress/DonutProgress.java
@@ -8,23 +8,29 @@ import android.graphics.RectF;
 import android.view.View;
 
 /**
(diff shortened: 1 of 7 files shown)


QuantumBadger:
This looks good, thanks! Merging it now.

PR #650, opened 2019-02-01
Allow enter key/action button to always be used to initiate search.

Cguy7777:
Currently in RedReader the enter key can be used to start a search from the main menu, but not from a subreddit or comment search or from the header of a previous subreddit search.


QuantumBadger:
Thanks!

PR #665, opened 2019-05-08
Update string.xml

Tuxicoman2:
Fix wrong FR translation

### DIFF
--- a/src/main/res/values-fr/string.xml
+++ b/src/main/res/values-fr/string.xml
@@ -126,7 +126,7 @@
 
     <!-- Main Menu -->
 
-    <string name="mainmenu_frontpage">Front Page</string>
+    <string name="mainmenu_frontpage">Page d'Accueil</string>
     <string name="mainmenu_all">Tous les subreddits</string>
     <string name="mainmenu_custom">Subreddit personnalisé</string>
     <string name="mainmenu_custom_invalid_name">Nom de subreddit invalide.</string>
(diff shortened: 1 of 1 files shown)


src/main/res/values-fr/string.xml
```diff
@@ -126,7 +126,7 @@
 
     <!-- Main Menu -->
 
-    <string name="mainmenu_frontpage">Front Page</string>
+    <string name="mainmenu_frontpage">Page d'Accueil</string>
```
QuantumBadger: I think the build is failing, because this apostrophe needs to be escaped with a backslash:

```suggestion
    <string name="mainmenu_frontpage">Page d\'Accueil</string>
```

src/main/res/values-fr/string.xml
```diff
@@ -195,11 +195,11 @@
 
     <string name="action_menu_post_title">Action sur le post</string>
 
-    <string name="action_upvote">Upvote</string>
-    <string name="action_downvote">Downvote</string>
+    <string name="action_upvote">Vote positif</string>
```
QuantumBadger: Are the words "upvote" and "downvote" not used on the French reddit? They seem like reddit-specific words, rather than English words that need translating. I'm not the expert on that though :)

Tuxicoman2: I don't know what the other languages do. It's a bit strange to use english wording whereas there is a french translation for a positive vote or a negative one.

https://fr.bab.la/dictionnaire/anglais-francais/upvote

QuantumBadger:
Thanks, merging this in.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/6aaa3fb1a73e4e016fe66588234a9509b7ee7536

PR #674, opened 2019-07-05
Updated deprecated ClipboardManager

sbaraka:
android.text.ClipboardManager did not meet the minimum API level for the application. Updated it with current


QuantumBadger:
Thanks for the pull request, although even after this change the deprecated `setText()` method is still used, we're now just accessing it through a subclass (`android.content.ClipboardManager`) of the original deprecated class (`android.text.ClipboardManager`). I'll close this for now, but I'd accept a PR which uses the new API.

Issue #639, opened 2018-12-09
v.reddit videos save as jpg files


TippyLion28:
v.reddit videos save as jpg files

sharkwouter:
I've created pull request https://github.com/QuantumBadger/RedReader/pull/681 which fixes this issue.

PR #681, opened 2019-10-01
Save v.redd.it videos as mp4 when downloading

sharkwouter:
I'm not sure if this is the best approach here, but this does fix https://github.com/QuantumBadger/RedReader/issues/639

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/General.java
+++ b/src/main/java/org/quantumbadger/redreader/common/General.java
@@ -58,7 +58,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URI;
-import java.net.URL;
 import java.nio.charset.Charset;
 import java.security.MessageDigest;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


QuantumBadger:
Thanks, this looks good!

At the moment this just checks if the URL string contains the text "v.redd.it", but it might be better to save the result of `uriFromString()` on line 443, and call `getHost()` to get the hostname:

    final URI uri = uriFromString(url);
    ...
    if("v.redd.it".equals(uri.getHost()) {
    ...

If you'd like to make (and test) the above change please go ahead, or otherwise I can merge as-is.

sharkwouter:
I've made the requested change. It works just as well so I've added it to this pull request.

sharkwouter:
Okay, now it actually passed the pmd test. My bad.

QuantumBadger:
Perfect, thanks! I'll get this merged in.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/f594cd4010ff7206c169e8f0d1c98b5a6f3a387a

Issue #637, opened 2018-12-04
Allow copy text from a post main body.


nathanernest:
Comments can be coppied with a long press and selecting "copy text" but no such option is available for the main body of text from a post. Would be a great feature to have.

Efreak:
There's a confusingly-labeled 'copy' item when long-pressing the post title, but it only copies the post url, not the contents. This *should* be renamed 'Copy post URL', or should be configurable in options (choose from url, title, content, or some combination)

To copy text post, reply to post, then long-press the original text in the lower half of the screen to get a selection. This *should* be available in the main post view as well.

*Should* meaning these are things the app does not do, but seem like good ideas to me. *[Shoooood](https://en.m.wikipedia.org/wiki/Semantic_satiation)*

MichalFerencsik:
Hi @QuantumBadger , can I work on this issue?

japanesephundroid:
> Hi @QuantumBadger , can I work on this issue?

Not QuantumBadger but he has said repeatedly that he welcomes pull requests as he does not have much time for the app now. He just pretty much only looks at pull requests and fixes critical bugs that break the app.

As this seems like a small fix and not a major feature that fundamentally changes the app, I doubt that he would have any objection to it.

MichalFerencsik:
Thanks for your reply.
Pitty is that I already prepared change for this. 😌

Nevermind, I will take a look for something else. 🙂

MichalFerencsik:
Also, @japanesephundroid, do you think that Issue#667 would be worth to try?

japanesephundroid:
>Thanks for your reply.
>Pitty is that I already prepared change for this. 😌
>Nevermind, I will take a look for something else. 🙂

Just submit your change, I don't know what you understood from my reply put he will look at it eventually.

> Also, @japanesephundroid, do you think that Issue#667 would be worth to try?

Yes. It's a small change I think.

QuantumBadger:
@MichalFerencsik Yes, please do feel free to submit a pull request and I'll take a look at it.

QuantumBadger:
Thanks for the proposed change @MichalFerencsik, I've left a comment on the PR.

In the meantime, to copy self-text it's possible to long click the post title and select "properties". At the bottom of the properties dialog is the self text, which you can highlight and copy.

Cguy7777:
@QuantumBadger This seems to have been implemented. :-)

QuantumBadger:
Thanks @Cguy7777, closing :)

PR #682, opened 2019-10-01
Issue#637: Implement copy to clip body of post

MichalFerencsik:
Issue: #637 Allow copy text from a post main body.

Description: Implement copy of post body to clipboard using OnLongClickListener in _CommentListing_
activity.


QuantumBadger:
Thanks!

Normally when long-clicking in RedReader you'd expect the long-click context menu to appear. I think it might be better to (1) show the context menu when long-clicking (which is what already happens if you long-click the post title, above the post body), and (2) add an item to the context menu called "Copy self-text". This would be more in line with what people expect.

Let me know what you think :)

MichalFerencsik:
Hi @QuantumBadger, 

**(1) show the context menu when long-clicking (which is what already happens if you long-click the post title, above the post body)**
_Can we resolve this by opening same dialog when user makes a long click on post itself?_

 **(2) add an item to the context menu called "Copy self-text". This would be more in line with what people expect.**
_Done, yet it would require migration, and I'm not sure about manifest change_

QuantumBadger:
Great work, thanks!

This basically looks ready to merge, although I think there's a minor bug -- both "Copy" and "Copy Self-Text" seem to copy the self text? The first one used to copy the URL:

https://github.com/QuantumBadger/RedReader/pull/682/files#diff-efdaf8c90823a78e85dc5c0bf0c4fcfcR471

>  Can we resolve this by opening same dialog when user makes a long click on post itself?

Yes I think that's the right approach.

>  Done, yet it would require migration, and I'm not sure about manifest change

Your change looks fine to me, if you have any questions please let me know :)

MichalFerencsik:
I reverted minor bug inserted by copying code due to update of Clipmanager. 

Is it ok now? 

Thanks,
Michal

MichalFerencsik:
Hi @QuantumBadger, 

have you had time to look at this?

Thanks,
Michal

QuantumBadger:
Thanks @MichalFerencsik, it looks fine. Merging in now!

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/c5d49e2491df51cb0dfebdf6de581fc2ffa0bdd7

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/3c4ff6211237f91145e710a9c57b1eb2893a7722 (correct commit this time :))

Issue #550, opened 2017-12-05
"Skip to front page" setting ignores "Tablet mode" setting


vjon:
If the "Skip to front page" setting is enabled, the app will open to the front page in a Single pane, even if:
- "Auto-detect" is selected for tablet mode and the device is a tablet-class screen size
- "Always" is selected for tablet mode

Pressing the back button to return to the main menu then activates dual-pane view as expected, until the next time the app is closed and re-opened.

Tested on the following devices:
- Google Pixel 2 running Oreo 8.0.0 ("Always" tablet mode option used for testing)
- Google Pixelbook running Chrome OS 62.0.3202.97 (Both tablet mode options used for testing)

Cguy7777:
@QuantumBadger Looks like this has been fixed.

PR #684, opened 2019-10-03
Fix "Skip to Frontpage" disregarding two-panel option

LizardWithHat:
This could fix issue #550. It was as easy as setting two panel mode and updating before executing the skipping preference on App startup.
The app now behaves as expected, opening in the two-panel mode while instantly loading the frontpage results in the right panel.
I tested this with:
- Nexus 7 2012 (small Android 7 Tablet device)
- Virtual Android 5 Phone
- Virtual Android 8 Phone
- Virtual Android 9 Tablet


QuantumBadger:
Excellent, thanks! I'll merge this in.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/b176f4c6eda54091fd375574caa1a4e4d45e1859

PR #686, opened 2019-10-10
Introduce built-in Share Dialog with prioritization

### LINKED ISSUE #581: [Feature Request] Improve 'share' feature by adding quick access to 3 applications last shared to
Take a look at how Firefox does it - very comfortable.
I use 'Telegram' IM and 99% of the time I share links to it, but since the list of applications to share to is sorted alphabetically - I have to scroll down every time.

Another approach would be to just sort the list differently: make items get listed in the most recently used order.

LizardWithHat:
As mentioned earlier this month, I somewhat finished implementing an own Share Dialog. This should close #581 if it suffices.
The new share dialog reuses the list_item.xml to build a List of Applications that can share the selected content and prioritizes three recently (within RedReader) used Apps.
The edge-cases of having less than three and no available Application to share the content with are handled.
This solution as is should be able to handle a variable amount of prioritized Apps and can be changed through the static Int in ShareOrderDialog. That's something I could imagine to be a configurable option, too.

I'm aware that doing this with that many Lists is not performant, though there is no way to persist the ResolveInfo without relying on external library like GSON. This is why I persist a concatenated String with the Intent names and go through the list, pulling out the prioritized Apps through their Intent name and later re-add them in their respective order to the beginning of the list.
I'm open for suggestions on how to improve it.

Here's an example picture;
image

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/WebViewActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/WebViewActivity.java
@@ -20,6 +20,7 @@ package org.quantumbadger.redreader.activities;
 import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
@@ -135,8 +136,14 @@ public class WebViewActivity extends BaseActivity implements RedditPostView.Post
 						mailer.putExtra(Intent.EXTRA_SUBJECT, mPost.title);
 					}
 					mailer.putExtra(Intent.EXTRA_TEXT, currentUrl);
-//					startActivity(Intent.createChooser(mailer, getString(R.string.action_share)));
-					ShareOrderDialog.newInstance(mailer).show(getSupportFragmentManager(), null);
+
+					if(PrefsUtility.pref_behaviour_sharing_dialog(
+							this,
+							PreferenceManager.getDefaultSharedPreferences(this))){
+						ShareOrderDialog.newInstance(mailer).show(getSupportFragmentManager(), null);
+					} else {
+						startActivity(Intent.createChooser(mailer, getString(R.string.action_share)));
+					}
 				}
 				return true;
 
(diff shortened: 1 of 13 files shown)


QuantumBadger:
Thanks, and sorry for the late response!

This looks good, however:

* I think the Android share dialog (on my phone at least) is nicer, and e.g. remembers contacts too. Maybe worth using the system dialog by default on Android 8.0+?

* The icons in the list are different sizes, although I guess this depends on the apps you have installed:

screenshot_share

LizardWithHat:
> I think the Android share dialog (on my phone at least) is nicer, and e.g. remembers contacts too. Maybe worth using the system dialog by default on Android 8.0+?

That's what I observed, too. I can't tell when Android added that, but it's been like that for a while. It might be worth setting the native Share Dialog as the default for all versions. What do you think?

> The icons in the list are different sizes, although I guess this depends on the apps you have installed

Yes, that's most likely because I reused the list_item.xml which doesn't restrict the image size. I think a new layout file with that restriction should do it. Just wanted to get this idea out for discussion as soon as possible.

And don't worry about responding late, I've been busy myself.

QuantumBadger:
> It might be worth setting the native Share Dialog as the default for all versions. What do you think?

Yeah that seems reasonable, people can enable it if the share dialog on their system is the "old" one.

> Yes, that's most likely because I reused the list_item.xml which doesn't restrict the image size. I think a new layout file with that restriction should do it. Just wanted to get this idea out for discussion as soon as possible.

Thanks, sounds good.

LizardWithHat:
So, I fixed a little oversight while I was at it.
The native share dialog is now the default one. I put option under "Behaviour" to toggle it.
The new layout file now limits the icon sizes and behaves correctly.

Screenshot_20200111-140431_RedReader

QuantumBadger:
Thanks, I've tried this locally and it looks good!

Before I can merge this, please could you add copyright headers to the new Java files? Once that's done I'll get this merged in :)

LizardWithHat:
Just did that. The layout file is a copy from the item list one, so it already has the copyright header.

QuantumBadger:
Thank you! I've now merged this in and updated the changelog.

PR #688, opened 2019-11-02
first attempt at creating tables for RedReader

Syntax3rr0r:
It has been a while since I used java or any OO Programing language. So apologies for the bad code.
my main issue is that RedReader is not using mono spaced font so the table is bound to be off. 
I just hope that this doesn't end up being more work to fix than write.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/reddit/prepared/markdown/MarkdownParser.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/prepared/markdown/MarkdownParser.java
@@ -153,7 +153,7 @@ public final class MarkdownParser {
 			//We are expecting the 2nd line to be the alignment if it does not exist then it
 			// is an invalid table so we ignore it.
 			// we do this by a REGEX ofc.
-			if (currTable[1].toString().matches("^([-|:]|\\|:)-*:?\\|(?!$)(:?-+:?\\|?(?!$))*:?-+:?\\|?$")) {
+			if((currTable.length > 2) && (currTable[1].toString().matches("^([-|:]|\\|:)-*:?\\|(?!$)(:?-+:?\\|?(?!$))*:?-+:?\\|?$"))){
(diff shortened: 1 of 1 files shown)


QuantumBadger:
Thanks for the pull request, sorry I didn't respond sooner!

I've tested this locally and it doesn't seem to do anything? For example, when viewing [this user account](https://old.reddit.com/user/decronym), the tables look no different to before.

> RedReader is not using mono spaced font so the table is bound to be off

Yeah, I think what we need to do is maybe render tables in a separate activity (after clicking a "View table" button), possibly by converting them to HTML and rendering them in a WebView (taking into account any markdown inside the table).

Syntax3rr0r:
sorry for the late follow up
I agree that it nothing much. but my goal was to get some foundation for the table parsing. my fear is that it is currently very badly written and I am out of touch with java and not familiar with the objects in RedReader.
one last note is: it is different as previously it used to take all of the lines for the table and spit them out as a single line. which made reading anything much harder.
as for my approach for the markdowns it was yet another hot mess as I am still not sure why you have the objects being a start and size of line the character stream.
made my original plan of injecting the table after formatting impossible. my only request at the moment if you dont mind trying to work with me in trying to find a proper place for the code. as rn it sticks out like a sore thumb and I am ashamed of it.
also how would you suggest working on parsing an HTML in redreader as again I am very lacking in the inner working of it.
thank you for time!

Syntax3rr0r:
yeah clearly I am need to better understand how RedReader works, as just like you said it fails at the comments spectacularly might I add example here:
https://www.reddit.com/r/YouShouldKnow/comments/y37p6/ysk_how_to_make_a_table_on_reddit/
it fails at parsing altogether!
will look into it by weekend at the latest

QuantumBadger:
Closing for now due to lack of activity. Let me know if you want to have another go in future -- I'm happy to help out, although this is a pretty difficult issue and there are probably easier things to get started with.

Issue #680, opened 2019-09-07
[feature request] highlight which sorting of post/messages is being used


WPFilmmaker:
I am using Night (low contrast) and when I sort post/messages there is no way to see which one is being used, unless you scroll to see the date or the points, which could be easily made handier if the next time I click on the '3 lines' button for sorting, the current sorting order is highlighted.
Thank you.

RedReader v. 1.9.10.

PR #690, opened 2019-11-23
Highlight which sorting of post/comments is being used

KelianB-IMTA:
Implements the feature detailled in issue #680 
The currently selected sorting method when displaying posts or comments is now highlighted in red.


QuantumBadger:
Thanks, and sorry for the late response. I've tried this out and there are a few issues:

* The "Hot" option is always highlighted on `/r/all`, even if a different sorting method is selected
* When using the Ultra Black theme, the main menu is a lighter colour than before
* The "Top: All Time" order has been removed
* When a Top sort is selected, the word "Top" isn't highlighted in the first menu
* If possible, it would be better to use radio buttons rather than highlighting in red: https://i.imgur.com/FaIeeS8.png

If you can fix the above I'd be happy to merge this in :)

QuantumBadger:
Closing this due to lack of activity -- please re-open if the above gets fixed!

PR #694, opened 2019-12-16
added draft feature on posts

### LINKED ISSUE #557: request: automatically save drafts of self-posts
more than once i wrote a long post or made an edit, and then accidently pushed back (ghost moves on my touchscreen) and the whole edit or post is lost

thread on reddit: https://www.reddit.com/r/RedReader/comments/7n1d1x/request_automatically_save_drafts_of_selfposts/

fguibourt:
**Draft feature for postsubmit. Fixes #557**
This feature makes it possible to save a draft for a post with static variables.
The changes are in PostSubmitActivity.java. There are mainly based on the draft functionnality in CommentReplyActivity.java.
The fields stored with static variables are :

- type of post (String)
- post subreddit (String)
- post title (String)
- post content (String)
- if the replies should be sent to inbox (boolean)
- if the post is NSFW (boolean)
- if the post is spoiler (boolean)

**Before :** when a user wrote a post on a subreddit, this post was lost if the user went back to the subreddit page or to another page.

**Benefits :** when a user writes a post on a subreddit, this post is saved for the current subreddit. If the user go to another page, he can get its post back when he hits "Submit Post" on the same subreddit page.

**Limits :**

- If the user goes to another subreddit and hits "Submit Post", then the draft of the previous subreddit is lost.
- If the user quits the application, then the draft is lost.

**Next :**

- Maybe we could use a data class instead of static variables. Using the subreddit as a key, we could save a draft for each subreddit.
- Maybe suggesting draft saving as an option to the user. Indeed, saving drafts automatically can be annoying for some users.
- Use persistence to store the drafts even if the user quits the application.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/PostSubmitActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/PostSubmitActivity.java
@@ -65,6 +65,14 @@ public class PostSubmitActivity extends BaseActivity {
 	private static final int
 			REQUEST_UPLOAD = 1;
 
+	private static int lastType;
+	private static String lastTitle;
+	private static String lastSubreddit;
+	private static String lastText;
+	private static boolean lastNsfw;
+	private static boolean lastSpoiler;
+	private static boolean lastInbox;
+
 	@Override
 	protected void onCreate(Bundle savedInstanceState) {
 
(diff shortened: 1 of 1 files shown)


QuantumBadger:
Thanks, this looks good! One minor comment before I merge, it might be worth removing the "Draft saved" toast for now, as users might think it is saved to disk when actually it's just in RAM :)

fguibourt:
Done ;)

QuantumBadger:
Thanks! I'll merge this in.

I've noticed a minor issue where (on my device at least) the draft is still saved ever after successful submission -- I think this is because `onDestroy()` gets called after `resetDraft()`, and the same issue affects the existing comment submission code.

I'll merge this in now and then put in a quick for the above.

QuantumBadger:
Issues fixed and changelog updated :)

PR #695, opened 2019-12-16
Solves issue 637

Khaliiil:
This can solve the "copy titles" problem


QuantumBadger:
It looks like this breaks the "copy self text" feature by copying the title instead?

PR #702, opened 2020-02-03
Implement finer control over font scaling

Cguy7777:
The other day I thought that people might like it if you could change the font sizes of post titles and post subtitles separately, and I went from there.
I also noticed that post fonts were not being changed when being shown as a header, so I added separate options for posts in subreddit listings and for post headers. I'm not sure there's much reason to want those to be different, but I figured that it's always a good idea to give users the option of retaining the current behavior.


Cguy7777:
As of right now I have it set up so that the current post header behavior (staying at 1x regardless of post font scale) is retained when updating, but not the default on new installs. However, I think it might be better to also use the new-install default when updating, where the post header font scale is the same as the normal post font scale. I imagine that most users who've adjusted the post font size would like it to be that way in headers too. Leaving it as is may prevent users unhappy with the current behavior from discovering the new option, and those wishing to retain the current behavior could simply set it back.

QuantumBadger:
Good work, thanks! I've merged this in.

One quick suggestion -- there are now a ton of font scale options, and I think it would be good to move these to their own settings submenu ("Font") alongside the existing "Appearance" submenu. The names of the preferences could also be tweaked to make it clear they're for scaling (e.g. "Self text" should be "Self text scale").

QuantumBadger:
Changelog updated.

japanesephundroid:
Thanks for this.
Some notes.

I always set all font sizes to 1.5x.

When I do the same on this (including the new ones) the post title/subtitles are much bigger when I enter a post (at around 2.0x).
I can fix this by ticking **Use different settings for opened posts** and  selecting **Normal** for both the available options. As far as I can tell that is not intended behavior.

Also after doing all the above I have **Show Link Buttons** ticked and the title text size for the created button is again bigger than it should be.

Cguy7777:
@japanesephundroid, I looked at the font sizes and you are correct, the size of the text is larger in opened posts (at 1x, 19 & 13 sp in opened posts vs 14 & 11 sp in subreddit listings for titles/subtitles).

When I wrote this PR I simply multiplied the larger fixed sizes by the font scale preference, since the (sub)title in opened posts did not scale at all before. You can make them roughly equal by using a smaller scale for them (as you did), but perhaps now that they can be scaled they should be set to the same default size?

Unfortunately if we changed it to use the listing sizes it would be impossible to get the _exact_ same current sizes back, but I don't think most people would notice the difference.
(The closest to the current title size would be 19.6 sp at 1.4x; for subtitles, 13.2 sp at 1.2x)

On the other hand, the scaling preferences are meant to scale relative to the default sizes of things, and I don't think it's inherently unreasonable for the post (sub)title to be larger than normal when you've opened up a post, so maybe it should be left as is? What do you think @QuantumBadger?

-----

As for the link buttons, those were also not scaled at all before and now use **Comments Font Scale**, which is why they're larger, but on second thought they should probably have their own preference. Doing so would make the scaling preferences even longer (nine preferences!), but as [QuantumBadger mentioned](https://github.com/QuantumBadger/RedReader/pull/702#issuecomment-589939480) those should probably be moved to their own page either way.

Cguy7777:
On the subject of the **Font** section, there are, as of right now, eight different preferences plus the checkbox, and it looks like a good idea to add another one. Even with the checkbox disabled, requiring users to modify six or seven settings to scale everything seems excessive.

Instead of just that checkbox, perhaps there could be a ListPreference with different degrees of customization? It could maybe look something like this:

**Font scale customization mode**
- Simple [One preference for scaling everything]
- Advanced [3 preferences: Posts (controlling all titles/subtitles and self text), Comments (controlling  comment headers, text, and link buttons), and Inbox]
- Full [All preferences]

japanesephundroid:
>As for the link buttons, those were also not scaled at all before and now use Comments Font Scale, which is why they're larger

Then why is only the title larger, and the link text smaller? Also I noticed after posting that when the button has no title then the link text font size is larger.

I also think **Fonts** should have its own section, a list preference would be a good compromise.

QuantumBadger:
> Instead of just that checkbox, perhaps there could be a ListPreference with different degrees of customization?

It's a good idea, although rather than having three "levels" like this, I think it might be simpler to have:

* One "global font scale" preference (at the top of the new "Fonts" submenu), which affects all text in the app
* The eight (or nine) detailed preferences underneath it, with each of their default values set to "Use global scale"

Users then have one preference to scale everything, but can override specific things if they want. Other tweaks:

* We could get rid of the checkbox ("Use different settings for opened posts") for simplicity
* I think the "Self text", "Comments", and "Inbox" preferences should just be combined into one. They're all the same "type" of text (long-form markdown). The preference could just be called "Comment text scale".

Hopefully that's clear, please let me know if you have any thoughts :)

PR #703, opened 2020-02-04
Record lastVersion on first application launch

Cguy7777:
While I was testing #702 I discovered that RedReader does not record its current version into the preferences file on its very first launch, which awkwardly results in the changelog being shown on the app's second launch even if the app hasn't been updated.
More importantly, it means that if an update that performs a one-time change to make new preferences match the user's current settings is installed, the change will not be performed.
This PR will prevent this on all future releases after its implementation.


QuantumBadger:
Well spotted, thanks! Changelog updated.

PR #705, opened 2020-02-06
Change several preferences to ListPreference's

Cguy7777:
I saw a TODO in there to change the "Precache images" settings to a ListPreference, so I thought I'd tackle that, as well as changing "Precache comments" and "Show thumbnails", since they seemed to be equally deserving of the change. Making "Show NSFW thumbnails" stay disabled when the new thumbnail setting was set to Never was surprisingly challenging.


QuantumBadger:
Looks good, thanks! I've merged this in.

PR #706, opened 2020-02-08
Show error and stop loading NSFW subs if NSFW content is disabled

### LINKED ISSUE #691: [feature enhancement] when nsfw subs do not load because the option is not ticked, no proper error message is displayed
### Steps to reproduce
1. Open RedReader
2. From the welcome page, click on "Custom Location"
3. Pick "Subreddit" in the dropdown list (should be the default choice)
4. Write the name of the sub you want to visit. Pick a "not safe for work subject". 
Here, I picked"r/gonewild"
5. Press go

### Expected behaviour

The sub should display an Error message such as "This subreddit is flaged nsfw. If you wish to display this content, enable it from the Settings > Behaviour section. "

### Actual behaviour

The sub loads indefinitely and never fully displays.
- After refreshing
Capture d’écran 2019-12-04 à 11 34 37
- Before refreshing (using cache)
Capture d’écran 2019-12-04 à 11 34 26

Cguy7777:
This should make it more obvious what's happening when RedReader won't load NSFW Subreddits. Since there's no point in downloading the data if every post is guaranteed to be blocked, this should also stop requests after getting the Subreddit's info.
Closes #691

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
+++ b/src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
@@ -280,8 +280,17 @@ public class PostListingFragment extends RRFragment
 									@Override
 									public void run() {
 										mSubreddit = result;
-										onSubredditReceived();
-										CacheManager.getInstance(context).makeRequest(mRequest);
+
+										if(mSubreddit.over18 && ! PrefsUtility.pref_behaviour_nsfw(context, mSharedPreferences)) {
+											mPostListingManager.setLoadingVisible(false);
+											mPostListingManager.addFooterError(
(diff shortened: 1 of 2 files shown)


QuantumBadger:
Thanks! The lack of information was intentional originally, due to Google banning reddit apps for allowing access to NSFW subs:

https://www.androidpolice.com/2012/03/19/reddit-is-fun-gets-banned-from-the-google-play-store-for-sexually-explicit-material-reddit-users-in-an-uproar/

I think that's less of an issue now, and in principle I'm happy to show more information to the user (not sure whether to mention the option in settings though).

QuantumBadger:
Now merged, after changing the text a bit. I decided that it's probably safe to mention the settings :)

PR #707, opened 2020-02-13
Return to origin from side toolbar

### LINKED ISSUE #677: Leaving post image opened from comments view via slide toolbar leaves post
1. Open comments view on an image
2. Scroll down to some comment (not really necessary, just keep it in mind)
3. Open the slide menu to view the post image
4. Hit back button or tap screen to close the image
5. Instead of going back to the comments view, the entire post has now closed, leaving you back in the previous screen. If you weren't collapsing comments as you read them, it's now difficult to find your place in a long thread.

Cguy7777:
This should fix #677. I'm not very familiar with activities, but it looks to me like the activity you came from should be finished by simply going back again once you return to it. But maybe I'm wrong and this would cause memory issues?
Closes #677

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java
@@ -542,12 +542,10 @@ public final class RedditPreparedPost implements RedditChangeDataManager.Listene
 				break;
 
 			case COMMENTS_SWITCH:
-				if(!(activity instanceof MainActivity)) activity.finish();
 				((RedditPostView.PostSelectionListener)activity).onPostCommentsSelected(post);
 				break;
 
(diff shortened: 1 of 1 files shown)


QuantumBadger:
Thanks! I think #677 was intentional behaviour, that if you tap that button it replaces the current activity with the other one (i.e. if you're viewing the comments, it closes the comments and shows the image instead, and vice versa).

The problem with the fix in this PR is that, if you repeatedly swap between them with the button, the number of activities will keep increasing, and you'll have to go back through all of them one by one:

https://streamable.com/0z29h

Ideally the behaviour of the button would be as follows, when switching from image to comments (and the same in reverse):

* Check the stack of activities to see if the comments are already open in the background
** If so, swap the position of the two activities, bringing the comments to the front
** If not, open a new activity displaying the comments.

Not sure how easy Android makes it to do this!

PR #708, opened 2020-02-13
Use strings on Search Results

Cguy7777:
This should allow us to translate the words used on the search results page.
Since there are two strings to format in if both the query and the Subreddit are being printed, I added a string to toggle whether the query or the Subreddit is used first in String.format, in case printing the Subreddit first flows better in the given language. I used a \<string\> instead of a \<bool\> because I think only \<string\>'s will appear on Weblate, and the option should be easily available.


Cguy7777:
I forgot that argument indexes exist&hellip;

QuantumBadger:
Thanks, merged this.

PR #710, opened 2020-02-17
Some minor settings changes, mainly to fit Material Design

Cguy7777:
The main purpose of this PR is to change the preferences strings to match [Material Design's recommendations](https://material.io/design/platform-guidance/android-settings.html), by changing capitalization and re-wording several strings. It might be a good idea to add a link or note about this to the [wiki page about adding preferences](https://github.com/QuantumBadger/RedReader/wiki/%5BDev%5D-Adding-a-Preference) to help keep the strings for new preferences in line.

This PR also has a few other minor changes: 
- Make the **Use Custom Tabs** preference dependent on **Use internal browser**, since it has no effect if that's disabled
- Move **Tablet mode (two pane)** to the top of the appearance preferences and remove its PreferenceCategory (it was the only one in there)
- Add a summary for **Use Tor** to remind users that links will not be sent over Tor if they are not using the built-in WebView browser
- Add a summary for **Comment header** to clarify its purpose

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -37,14 +37,13 @@
     <!-- Appearance Prefs -->
 
     <string name="pref_appearance_theme_header">Theme</string>
-    <string name="pref_appearance_font_header">Font</string>
+    <string name="pref_appearance_font_header">Font Scale</string>
     <string name="pref_appearance_thumbnails_header">Thumbnails</string>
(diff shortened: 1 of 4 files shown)


QuantumBadger:
Merged in, thanks!

https://github.com/QuantumBadger/RedReader/commit/edcfd737441af1af8c2d7f90c3634b0b8aa22d33

PR #711, opened 2020-02-20
Respect browser preference in internal WebView's

### LINKED ISSUE #491: Links in sidebar always open with internal browser, even if disabled
I have disabled the internal browser but links in sidebars seem to still use it.

Cguy7777:
While I was in there I also moved the code for opening a custom tab to its own function.
Closes #491


QuantumBadger:
Thanks, I've merged this in!

PR #716, opened 2020-02-28
Frame step for video player

japanesephundroid:
Frame step for video player. Exoplayer does not have a frame step option so this a frame step for a 30 fps video

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/views/video/ExoPlayerWrapperView.java
+++ b/src/main/java/org/quantumbadger/redreader/views/video/ExoPlayerWrapperView.java
@@ -146,6 +146,18 @@ public class ExoPlayerWrapperView extends FrameLayout {
 						}
 					}), buttons);
 
+			addButton(createButton(
+					context,
+					mControlView,
+					R.drawable.ic_action_back_dark,
+					new OnClickListener() {
+						@Override
+						public void onClick(View view) {
+							mVideoPlayer.seekTo(mVideoPlayer.getCurrentPosition() - 33);
+							updateProgress();
+						}
+					}), buttons);
+
 			{
 				final AtomicReference<ImageButton> playButton = new AtomicReference<>();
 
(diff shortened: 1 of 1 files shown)


japanesephundroid:
This is the first time I ever tried a pull request.

Don't know if its the correct way, sorry if not.

The icon I chose is not the best, but short of adding a new one it was the best already available in the app.

QuantumBadger:
Thanks! This looks good to me, when I get a chance I'll test/merge it and maybe add new icons.

I think it would be good to keep the new buttons hidden by default and have a preference to enable them, but I'm happy to do this if it's not clear how.

It might also be worth greying out the buttons while the video isn't paused (which may require another set of icons in a different colour).

japanesephundroid:
Thanks.

>It might also be worth greying out the buttons while the video isn't paused (which may require another set of icons in a different colour).

Maybe a better idea would be to add a pause to the frame step action if the videos isn't paused? I wasn't sure how to do it so I just submitted this.

I also wanted to add a speed button, something like 0.25x, 0.5x, 1x, 2x, 4x. Would that be difficult with Exoplayer?

One other thing would be current and total playing time somewhere on the video, but that is beyond my capabilities, I'll maybe do a feature request on that one.

codeofdusk:
@QuantumBadger, @japanesephundroid this PR has been open for almost three years without a review. Are there plans to merge it, or should it be closed as abandoned?

QuantumBadger:
@codeofdusk Yeah it's deliberately still open as it needs some changes, but nobody's got around to it yet. I'd still like to merge this feature, it just needs to be put behind an off-by-default preference. There may also be a better Exoplayer API for doing a frame step, rather than the current hardcoded timeskip.

japanesephundroid:
> @codeofdusk Yeah it's deliberately still open as it needs some changes, but nobody's got around to it yet. I'd still like to merge this feature, it just needs to be put behind an off-by-default preference. There may also be a better Exoplayer API for doing a frame step, rather than the current hardcoded timeskip.

This is another way to do it from the Exoplayer Github:
[997620](https://github.com/google/ExoPlayer/issues/7025#issuecomment-595080750)

> The first I am using `ExoPlayer` to play a video and create a `MediaCodec` to extract all frames
> When touching on the `SeekBar` i use the `GLES` to render those frames via GLUtils.texImage2D and call pause the media
> When touch-up I call `ExoPlayer` to seek to the position
> When user click play I hide the frame draw before :)
> It's working for me :)) Thanks!

QuantumBadger:
Closing as @ecawthorne has submitted #1207

PR #719, opened 2020-03-06
Move and reorganize font settings

Cguy7777:
I moved the font scale settings to a separate page as suggested.

Other changes:
- Created a separate font scale setting for link buttons
- Combined the settings for **Comments**, **Inbox**, and **Self text** into one new **Body text** setting
- Created a global font scale setting which overrides all others by default
- Removed the **Use different settings for opened posts** checkbox
- Changed the base font scale for self text from 14f to 13f to match the comment body text, since the settings have been unified
- Reworked the update code in MainActivity for the new setup


src/main/java/org/quantumbadger/redreader/activities/MainActivity.java
```diff
@@ -322,31 +322,53 @@ public void run() {
...
-					sharedPreferences.edit().putString(
-							getString(R.string.pref_appearance_fontscale_selftext_key),
-							existingCommentSelfTextFontscalePreference
-					).apply();
-
-					sharedPreferences.edit().putString(
-							getString(R.string.pref_appearance_fontscale_comment_headers_key),
-							existingCommentSelfTextFontscalePreference
-					).apply();
+					final String existingCommentSelfTextFontscalePreference = Float.toString(PrefsUtility.appearance_fontscale_bodytext(this, sharedPreferences));
```
QuantumBadger, line 329: `Float.toString` might break here, because not all fractional numbers have a floating point representation (this may return something like 0.7999999999 instead of 0.8). We should get the actual preference string here rather than a float.

src/main/java/org/quantumbadger/redreader/views/LinkDetailsView.java
```diff
@@ -62,8 +62,13 @@ public LinkDetailsView(Context context, final String title, final String subtitl
 		layout.addView(textLayout);
 		((LinearLayout.LayoutParams)textLayout.getLayoutParams()).setMargins(0, marginPx, marginPx, marginPx);
 
-		final float linkFontScale = PrefsUtility.appearance_fontscale_comments(context, PreferenceManager.getDefaultSharedPreferences(context));
-		
+		final float linkFontScale;
+		if(PrefsUtility.appearance_fontscale_linkbuttons(context, PreferenceManager.getDefaultSharedPreferences(context)) != -1) {
```
QuantumBadger, line 66: It might be better to put this logic instead `appearance_fontscale_linkbuttons()` (and similar methods), because at the moment these if statements are scattered across the code, and it's good to avoid methods returning magic numbers (-1).

Cguy7777:
@QuantumBadger Thanks for the feedback, I've implemented the requested changes.

QuantumBadger:
Thanks, this is now merged! Sorry for the delay in looking at this.

I've fixed what I think is a bug in the version migration here, where the global font scale could have been set to `-1`: https://github.com/QuantumBadger/RedReader/commit/60d8f82ce81f37283ef5282d421f4bfd4642e507

Issue #733, opened 2020-05-13
Feature Request: redgifs.com support (from gfycat for migrating nsfw content there)


japanesephundroid:
See announcement:
https://www.reddit.com/r/gfycat/comments/g9qim4/adult_content_announcement_2/

Since May 12 all NSFW content are now expected to be posted there instead of gfycat.
Old NSFW posts are expected to migrate to the new site.

PR #736, opened 2020-05-24
feature: redgifs.com support

ticoombs:
This duplicates the gfycat support, for redgifs. 

But I am unsure how to "extend" the GfycatAPI class correctly. So I've left it as its own file, in case their API diverges 

Tested via building, and running on Android 10 Pixel

fixes: #733

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/LinkHandler.java
+++ b/src/main/java/org/quantumbadger/redreader/common/LinkHandler.java
@@ -52,6 +52,7 @@ import org.quantumbadger.redreader.image.DeviantArtAPI;
 import org.quantumbadger.redreader.image.GetAlbumInfoListener;
 import org.quantumbadger.redreader.image.GetImageInfoListener;
 import org.quantumbadger.redreader.image.GfycatAPI;
+import org.quantumbadger.redreader.image.RedgifsAPI;
 import org.quantumbadger.redreader.image.ImageInfo;
 import org.quantumbadger.redreader.image.ImgurAPI;
 import org.quantumbadger.redreader.image.ImgurAPIV3;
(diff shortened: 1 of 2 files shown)


jakern:
Since both sites should use the same api have you tried just adding to the gyfcat regex pattern to say 'or matches redgif patter '?

ticoombs:
I might be able to, but that's assuming it does not diverge in the future. The only real difference is the regex and the new endpoint. Which I could not figure out how to define without basically including the whole class.

` 		final String apiUrl = "https://api.redgifs.com/v1/gfycats/" + imageId;` 

I have not done java coding for a while. I'll see if I can make it nicer.

ticoombs:
Doesn't look like I can. Because all of the classes are `static`, as well as the `Gfycat`. I can't extend it and set the apiUrl to override for redgifs. 

It might be possible but would end up being more code for changing them all to non-static typed classes.

QuantumBadger:
Thanks! I've merged this in.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/793d68762e4ca64eaae7cf9258a6a13e3ed08892

japanesephundroid:
Thanks for this 👍

Works ok for redgifs.com links.
Old Gfycat links now identified as NSFW however redirect to gifdeliverynetwork.com.

example (obviously NSFW, I went with something relatively tame)
https://gfycat.com/whirlwindreadycats which redirects to:
https://www.gifdeliverynetwork.com/whirlwindreadycats

I don't know if this needs a separate thread or should be posted here since the video is hosted on redgifs:
https://thumbs1.redgifs.com/WhirlwindReadyCats.webm

ticoombs:
😞 Shall put up a pr to handle that in a couple hours. Should be simple for adding gifdeliverynetwork to the regex for redgifs

japanesephundroid:
> 😞 Shall put up a pr to handle that in a couple hours. Should be simple for adding gifdeliverynetwork to the regex for redgifs.
> 
> Edit: its not as simple as I thought. Can you create an Issue, and I might try looking at it on the weekend. As this is probably an edge case which will sort itself out over the next week or two, once the uploaders move to redgifs.

I will create an issue. It is an issue I think for old links since its so easy to sort subreddits.
Try to sort any NSFW subreddit to top:alltime and you will see this happening a lot.

Issue #436, opened 2017-02-02
Gyfcat always opens as *.gif instead of webm/mp4


133794m3r:
I don't know if posters are retarded and purposefully posting the insanely large gifs or not. But even if they are RedReader should rewrite the url to zippy.gyfycat.com/rest_of_normal_url.{mp4,webm}

QuantumBadger:
Thanks for the report!

For https://gfycat.com/LargeLoneBlueandgoldmackaw, the GIF is smallest because:

* The resolution is tiny (541x304, vs 1080p for the mp4).
* It appears to be dithered with a smaller colour space.
* The mp4 is 60fps, but according to ImageMagick the GIF is only 20fps (`Delay: 5x100`).

I originally went with webm (including for imgur I think), but people complained about them not playing on their phones, which doesn't seem to be an issue with mp4.

If the webm files really are VP9 (as opposed to VP8), then that would partially explain the size difference, but I'm guessing that very few devices have hardware decode support for it (e.g. the Galaxy S6 has no VP9 hardware decoder, but the S7 does), if they're able to play it at all (Android 4.4 is the first version supporting VP9). I'd guess VP8 also has medicore support compared to H.264.

So I think mp4 is the most sensible default. Maybe a preference could be added though, to select the preferred codec.

QuantumBadger:
If we go with the preference, a better default might be `mobilemp4` for Gfycat:

https://thumbs.gfycat.com/LargeLoneBlueandgoldmackaw-mobile.mp4

133794m3r:
I said that that _sole_ one is the only one that's not using the default thing. Also the mp4 bitrate is always higher than the webm, and I _seriously_ doubt that it's vp9 vlc, and media info _both_ say that it's vp8.

Writing library                          : libebml v1.3.0 + libmatroska v1.4.0

Video
ID                                       : 1
Format                                   : VP8
Codec ID                                 : V_VP8
Bit rate                                 : 10.3 Mbps
Width                                    : 1 920 pixels
Height                                   : 1 080 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Variable
Compression mode                         : Lossy
Default                                  : Yes
Forced                                   : No


vlc says the following.


Codec: Google/On2's VP8 Video(VP80)

it _doesn't_ say vp9 so they're really vp8 and the bitrate is set so low likely because of that very reason that hardware acceleration is flakey at best on most of the devices out there.

[...]

PR #737, opened 2020-05-27
WIP: feature: Low bandwidth mode (gfycat/redgifs)

ticoombs:
Reasoning: 
Not everyone has fast internet, so people on slower networks, and smaller devices can benefit from the `.webm` files which are smaller in file size, but still contain the same quality.
Also for the people that utilize reddit over tor, this will result in a better user experience 
This poc has been completed and works, but 

UI Change:
- New menu item under Network, called "Low bandwidth mode"

My Thoughts:
- I'm unsure if adding another constructor is the right way to do this in `ImageInfo.java`. I'm up for some critique on it.
- Because we are now setting the image url in `ImageViewActivity.java` the cache is not shared across a toggle of `lowBwEnabled`. If you download something while `lowBwEnabled = True`, and then toggle it off, you will download the "high quality" version.
- `.webm` files save up to 80+% on file size. (one went from 88mb to 10mb)

Fixes: #436

----

Follow-up items/questions.

I just realized while creating the PR, that Saving & Sharing images as well as ImageInfo always show the `urlOriginal` var.

Currently, we don't have access to the user preferences in the ImageInfo & associated image apis. I followed the execution chain to one of the first places that already dealt with `PrefsUtility`. Unfortunately, for a correct implementation of a Low bandwidth mode, that means that we now need to change the below files, and possibly extend `PrefsUtility` into them, as anywhere that uses urlOriginal needs changing.

```
src/main/java/org/quantumbadger/redreader/image/SaveImageCallback.java
src/main/java/org/quantumbadger/redreader/image/ShareImageCallback.java
src/main/java/org/quantumbadger/redreader/fragments/ImageInfoDialog.java
src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
src/main/java/org/quantumbadger/redreader/adapters/AlbumAdapter.java
```

A possible plan of attack is to change everything from urlOriginal to `urlImage`, where its used in the app. Then in ImageViewActivity we set a new variable `urlImage` to either; `Original` or` LowBw`.

That way there would be standardized usage of the image's url across the app, if we are modifying it.

Thoughts?

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/ImageViewActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/ImageViewActivity.java
@@ -246,7 +246,9 @@ public class ImageViewActivity extends BaseActivity implements RedditPostView.Po
                 if (lowBwEnabled) {
                     Log.i(TAG, "Updated image URL: " + info.urlLowBw);
                     Log.i(TAG, "Updated image Type: " + info.typeLowBw);
-                    uri = General.uriFromString(info.urlLowBw);
+                    if (info.urlLowBw != null) {
+                        uri = General.uriFromString(info.urlLowBw);
+                    }
                 }
 
                 final URI audioUri;
--- a/src/main/java/org/quantumbadger/redreader/activities/ImageViewActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/ImageViewActivity.java
@@ -137,6 +137,8 @@ public class ImageViewActivity extends BaseActivity implements RedditPostView.Po
 		final int gallerySwipeLengthDp = PrefsUtility.pref_behaviour_gallery_swipe_length_dp(this, sharedPreferences);
 		mGallerySwipeLengthPx = General.dpToPixels(this, gallerySwipeLengthDp);
 
+        final boolean lowBwEnabled = PrefsUtility.network_low_bandwidth(this, sharedPreferences);
+
 		final Intent intent = getIntent();
 
 		mUrl = intent.getDataString();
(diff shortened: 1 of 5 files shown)


QuantumBadger:
Thanks for this, it's a good idea!

> New menu item under Network, called "Low bandwidth mode"

Rather than a single "low bandwidth mode" checkbox, it might be better to have two separate preferences for this, under "Images/Video":

* Use low-resolution images
* Use low-resolution video

These could then have three modes:

* Never
* Only on mobile data
* Always

By default, images could be set to "Never", but video could be set to "Only on mobile data".

> I'm unsure if adding another constructor is the right way to do this in ImageInfo.java. I'm up for some critique on it.

Yes I think this is reasonable, although the number of constructors there is pretty high now :)

It might be a bit neater to split out the three new fields (url, type, size) into a new `ImageOption` class. That way, `ImageInfo` would have two `ImageOption` members:

* `@NonNull public final ImageOption original;`
* `@Nullable public final ImageOption lowBandwidth;`

If so, you could just add a `@Nullable ImageOption` parameter to the existing constructor, rather than adding a new constructor. It would also make the code using the size/etc simpler, as you can just take a reference to the `ImageOption` you want, rather than having separate logic for original vs low bandwidth.

> Because we are now setting the image url in ImageViewActivity.java the cache is not shared across a toggle of lowBwEnabled. If you download something while lowBwEnabled = True, and then toggle it off, you will download the "high quality" version.

I think that particular case is fine. The reverse of that is less good, i.e. if you download a high quality image, then turn on lowBwEnabled -- visiting the image again would cause it to be downloaded a second time in lower quality, even though we have a high quality version in the cache.

In theory it's not too hard to fix:

* If low bandwidth mode is enabled, make a cache request for the high resolution version using `DownloadStrategyNever`, which will do a cache lookup but not download anything.
* If that fails, download as normal using `DownloadStrategyIfNotCached`.

However it becomes less easy in cases like this, where the actual download occurs in a different place to the decision:

https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/activities/AlbumListingActivity.java#L110

Maybe instances like that can be fixed in the future, after the basics of this are sorted out.

> Currently, we don't have access to the user preferences in the ImageInfo & associated image apis

That's true, although I think putting the decision in the files you mentioned (rather than e.g. inside `ImageInfo` itself) is the right approach.

Having a extra field inside `ImageInfo` could get a bit messy, e.g. if the preference changes then all the existing `ImageInfo` objects will have the wrong value.

> .webm files save up to 80+% on file size. (one went from 88mb to 10mb)

That's a very nice improvement! I know that a lot of people have requested this feature due to data caps, and it would be nice to finally add it.

If any of the above is unclear (or you have any more questions) please let me know!

QuantumBadger:
Closing this due to lack of activity -- please re-open if the above gets fixed!

PR #757, opened 2020-07-09
Improve navigation with TalkBack and directional controls

Cguy7777:
After spending some time testing RedReader with TalkBack, as well as directional controls, I've made a few improvements to navigation based on the recommendations (links below). These changes will go mostly unnoticed by touchscreen users.

The things that may be noticed:
- The addition of a **Shortcuts** heading at the top of the main menu. I put this in when marking the headings as such for accessibility services, because it would be a bit annoying to not have that section available when navigating by headings.

  I added a bit of code to reduce the top padding for that one heading, since I thought it looked a bit ugly having so much empty space between it and the very top of the app. Feel free to adjust or remove this.
- The ability to select text in the **Properties** dialog was removed because, unfortunately, I was unable to find a way to fix the focus for screen readers and directional controls without breaking it. It might be a good idea to add a new method for copying text from there (perhaps by long pressing on an entry), although it is probably not a major issue since you can copy some of that information from the context menu.

Recommendations I looked at:
- https://developer.android.com/guide/topics/ui/accessibility/principles
- https://developer.android.com/training/keyboard-input/navigation


Cguy7777:
On reflection, it struck me as silly to add an exception for the padding on just that one heading, especially when I realized that it's not the only heading that could be at the very top (if the user set their settings to show none of the shortcuts).

If it is considered too ugly as is, this could be accomplished in a better way, perhaps by adding a parameter for "Is this going to be at the top?" in **GroupedRecyclerViewItemListSectionHeaderView()** to trigger the padding change when necessary.

QuantumBadger:
Thanks for this! I haven't gone through the code changes in detail, but I have built+run it.

I'm eager to accept any changes which will improve accessibility, but unfortunately I'm not in favour of the two visible changes you mentioned above:

* The new heading wastes vertical space for the vast majority of users, and it doesn't look great (I don't think tweaking the padding will fix that). Is it possible to visually hide the text, but keep it logically present for the purposes of screen readers? I.e. users would see no visual change, but there would be an implicit "Shortcuts" header accessible by voice?

* Copying from properties dialogs isn't something I'd want to remove without a good alternative already in place. A long-press is reasonable as that's similar to what users already have to do -- maybe a long press could pop up a dialog, containing the value in a multi-line `EditText` control with something like `android:inputType="none"` and `android:textIsSelectable="true"`. Obviously be careful with long values, e.g. comment body text.

I'm happy to consider alternatives to the above though. Thanks again!

Cguy7777:
Messed up my branch, I'll make a new PR later.

PR #758, opened 2020-07-11
Improvements to copying to the clipboard

Cguy7777:
I made a few improvements to how copying is handled:
- Every instance of the deprecated **setText()** method was replaced with **ClipData**-based methods
- An informative toast is always displayed when copying
- The **Copy Self-Text** option in the post context menu is not shown if there is no self-text in that post
- Minor string changes for clarity

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/adapters/MainMenuListingManager.java
+++ b/src/main/java/org/quantumbadger/redreader/adapters/MainMenuListingManager.java
@@ -18,6 +18,7 @@
 package org.quantumbadger.redreader.adapters;
 
 import android.app.AlertDialog;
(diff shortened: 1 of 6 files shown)


QuantumBadger:
Looks good, thanks!

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/42f4355640e4f8b91082e96690acece5a996f014

japanesephundroid:
A small note/question.

Before this I could go to a comments properties and select which text I wanted to copy, now I can only copy the full text.
I found no way to enable this in settings.

Is it possible to get that ability back somehow?

PR #760, opened 2020-07-16
Improve navigation with TalkBack and directional controls

Cguy7777:
This is a redo of [my earlier pull request](https://github.com/QuantumBadger/RedReader/pull/757) due to utter incompetence with git.

After reviewing your recommendations, I made several changes:

- No heading is created for the **Shortcuts** section. Although I figured out a way to make one that is invisible visually but available to screen readers, it would look very jarring to see focus shift to nowhere for those who use screen readers but have some vision. I considered making an option for it, but it seemed a bit silly to create an entire preference over the display of a single line. Not having a heading at the very top is probably fine.

- The ability to copy text from each entry in the **Properties** dialog (and from every other `propView`) was re-added, by copying the text on a long press. This ability was also added for the karma counts in the **User Profile** dialog, for consistency. To avoid creating a large number of new strings and adding a new parameter to `propView` to figure out which toast to display, all copying done with this new method simply shows a generic "Copied to clipboard" toast.

##### Note
It's possible for there to be a heading at the very top if the user disables all of the **Shortcuts** items. I've attached a screenshot below to illustrate. It might be worth figuring out what heading is at the top (if any), and hiding it or reducing its top padding.

I looked into this, but it's a little more complex than I initially thought because you can't necessarily know at the start what headings will be displayed (e.g. if **Multireddits** would be at the top, you have to find out if the user has any multireddits to show).

Screenshot_1594881175

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/ChangelogManager.java
+++ b/src/main/java/org/quantumbadger/redreader/common/ChangelogManager.java
@@ -18,7 +18,12 @@
 package org.quantumbadger.redreader.common;
 
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.core.view.AccessibilityDelegateCompat;
+import androidx.core.view.ViewCompat;
+import androidx.core.view.accessibility.AccessibilityNodeInfoCompat;
+import android.os.Build;
 import android.view.LayoutInflater;
+import android.view.View;
 import android.widget.LinearLayout;
 import android.widget.TextView;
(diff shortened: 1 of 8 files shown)


QuantumBadger:
Looks good, thanks!

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/bf127cc42b09e9d263b3b2a1dd82b10c25ff319a

Issue #755, opened 2020-07-06
Simpler/faster account switching


alduraibi:
I switched to this app recently and I really like the work you've done. Thanks for all the work everyone put in!

A problem I noticed early on is the amount of taps it takes to switch between my two accounts. This becomes a hassle when you have to switch accounts frequently.

1. Tap options menu
2. Tap 'Accounts'
3. Tap on account name
4. Tap 'Make Active' (can easily click on delete by accident)

One suggestion is to switch accounts in step 3 and require a long press to get to a delete prompt. This way it at least lowers it to 3 taps and makes it harder to accidentally delete an account which is a decent improvement when switching accounts fairly often.

Another suggestion which I actually prefer a lot more is to have the alternative account/accounts listed directly in the three dot menu (either at the top or bottom). This way it reduces the number of taps to 2 which makes switching accounts feel a lot smoother. It also removes the possibility of accidentally deleting an account. You can even include this an option if you don't want to make it the default setting.

Issue #750, opened 2020-06-29
Preference labels do not wrap


Cguy7777:
The labels for RedReader's preferences are cut off when there is not enough space to display them on one line. This is most noticeable when the system font size is larger than normal.

Tested on both 1.10 and the current codebase ([latest commit when I tested](https://github.com/QuantumBadger/RedReader/commit/3fe213395cbf8f38b82db9806a84a82fc9884167)) on API 27 (example below) and API 16.

![Screenshot_1593405873](https://user-images.githubusercontent.com/44722506/85973999-f46adb80-b999-11ea-9af3-43bed75da918.png)

QuantumBadger:
Thanks! Looks potentially tricky to fix. The "right" way would be to migrate to `androidx.preference.PreferenceFragmentCompat` rather than using the deprecated APIs we currently use, although that's not a straightforward task.

PR #761, opened 2020-07-22
Allow app bar to be customized

Cguy7777:
After seeing some feedback on the number of presses it takes to switch accounts (e.g. #755), it occurred to me that it'd be great to be able to customize the entire app bar, so that users can have their often-used overflow menu items a single press away.

Every item in the app bar/overflow menu can now be set to one of four settings:
- Always show a dedicated button (This is higher priority than "Show button if room", but buttons will still be forced into the overflow menu if there is truly no space.)
- Show a dedicated button if there's enough room (This is good to have because the amount of space can easily change with a rotation, fold, etc.)
- Only show in the overflow menu
- Never show

The defaults for the new preferences match the current defaults (only **Sort** and **Refresh** as buttons, everything else except **Close All** in the overflow menu only), with one exception: by default for new users, **Accounts** is set to show if there's enough room. I think this would be a good default since it reduces the number of presses required to switch accounts, a common task for many. However, it is your call.

This is a work-in-progress, and has a few things to potentially address:
- **Search** (for posts) and **Search Comments** each show as separate, identical-looking buttons if the user has set **Search** to show a button and the user is viewing both posts and comments in two-pane mode.
    - Solution A: Create distinctive icons for each button. I can make those in Inkscape if needed.
    - Solution B: Make the **Search** button spawn a sub-menu with both options when this scenario occurs.
- Currently, a static image of an hourglass is used for the **(Un)Subscribe** button while (un)subscribing is in progress. It would be better to use an animated icon like a ButtonLoadingSpinnerView here (as is done in the recently-added post list header toolbar), although I'm not sure that's possible without a lot of work.
- Due to the number of individual preferences added, it would be strongly preferred to have the preferences for this new feature in their own PreferenceScreen. I've put them into one inside the Menu preferences, but it's a bit quirky: there is no app bar shown, and it takes the entire screen in two-pane mode. While this can be fixed by making it a top-level PreferenceScreen in the usual way, it seems ridiculous to make an entire top-level one just for this.
- It's worth thinking about which buttons users would likely use the most often and re-ordering accordingly, as [it is recommended](https://material.io/components/app-bars-top) that action items be ordered with the most-used icons on the left.


QuantumBadger:
Thanks, this is impressive work so far! Some initial comments:

* I think the static hourglass when subscribing is fine, no need to change it.
* In my brief testing so far, the "show icon if there's room" option doesn't seem to work consistently -- even if I set all the items to that, I never see more than 3 icons regardless of how much room is available.
* I'm undecided about whether it's a good idea to have "Accounts" as an icon by default, as it's potentially less discoverable for new users. Note that the first-run dialog for the app says "You can access the account list by selecting Accounts from the menu". Since the vast majority of users will have either one account, or no account, I'd keep it only in the menu by default.
* It could be confusing that the icon for "Accounts" is the same as the icon for "User Profile". Maybe consider `account-cog` or `account-details` from [here](https://materialdesignicons.com/)?
* On code style: there are a few places here where a fixed length array is created (e.g. `new OptionsMenuItemInfo[14]`) and then populated one element at a time over the next 14 lines (`optionsMenuItemsInfo[13] = ...`). It'd be neater to either:
  * Use `new OptionsMenuItemInfo[] { ... }` to automatically size the array based on the number of elements
  * Use an `ArrayList` (less efficient but doesn't really matter in this case)
  * Replace the `OptionsMenuItemInfo` class with an enum containing the 14 values, each having the two fields `OptionsMenuItemInfo` previously had
* Similarly, any arrays which are passed around that rely on an implicit ordering {e.g. the `int[]` returned by `pref_menus_optionsmenu_items`) should be avoided (e.g. replaced with a `Map`?)
* The preference key `pref_menus_optionsmenu_items_items_2` could maybe be renamed to be a bit clearer

> Solution B: Make the Search button spawn a sub-menu with both options when this scenario occurs

Yep, this sounds like a good solution to me.

> Due to the number of individual preferences added, it would be strongly preferred to have the preferences for this new feature in their own PreferenceScreen

Agreed, and I think the current approach of having this under Menus is good -- not sure why the action bar is missing though. If this ends up being in its own activity (even in two pane mode) that's fine by me.

If you wanted to look at migrating to `androidx.preference.PreferenceFragmentCompat`, that would be great, but also potentially a lot of work (related: #750)

> It's worth thinking about which buttons users would likely use the most often and re-ordering accordingly, as it is recommended that action items be ordered with the most-used icons on the left.

If you have any specific suggestions on this let me know, although for now I'm in favour of keeping this as-is.

Generally this is looking good though! Let me know if you have any questions.

Cguy7777:
@QuantumBadger I spent some time doing my best to refactor the code. I'm not 100% sure about all of it, so I'd love for you to take a look and tell me your thoughts.

In particular, after switching the preference method's return from an int[] to a Map, OptionsMenuUtility.prepare() is littered with warnings about potential NullPointerException's. I'm pretty sure it's impossible for anything from that Map to be null (except maybe if the preference storage is corrupted, but that would lead to crashes anyway), but would it still be worth adding logic to check for nulls anyway?

I also fixed a few bugs I found and thought about what the default for **Accounts** should be. On reflection, I think you're right: most users probably don't switch between accounts enough to justify having it as a button by default.

Cguy7777:
I believe I've addressed most of the things that needed to be, assuming the refactor is to your liking.

### Show button if room
After investigating, it appears that the behavior described is actually the intended behavior in Android.
Although it's not described in detail on the new page on material.io about app bars, I found [the former page](https://web.archive.org/web/20160127102230/http://developer.android.com/design/patterns/actionbar.html), which seems to still be correct, and states the rules for app bar buttons (presumably using SHOW_AS_ACTION_IF_ROOM) as:
  - Depending on screen width in dp's, the maximum number of buttons is from 2 to 5.
  - The buttons cannot take up more than half the app bar's width.
  - (The overflow button counts as one of the buttons.)

This looks pretty off when you have a lot of items set to "Show button if room", but I doubt many users will want more than several on screen at once.

Also, I think I made a mistake when I said that buttons set to "Always show button" would be put in the overflow menu anyway if too many of them were set to that and they really couldn't fit. Looking over it again, it seems that if too many buttons are set to always, they will simply consume all available space in order and leave no room for the overflow menu. This is probably not an issue, so long as the user is conscious of this when editing their preferences.

I think it is impossible for the user to lock themselves out of the preferences this way: based on the rules above, if buttons in "show if room" mode cannot consume more than half of the app bar, then you *should* always be able to fit in all four of the buttons on the Main Menu, including **Settings** (the smallest app bar would still have to be able to fit two buttons in half of it, and we don't care if we use up both halves).

### Button order
After looking over it, my tentative order for the buttons would be this:

1. Sort
2. Refresh
3. Search
4. Accounts (not sure if this should be before or after Submit Post & Reply)
5. Submit Post
6. Reply
7. Pin
8. Subscribe
9. Block
10. Sidebar
11. Past
12. Close All
13. Theme
14. Settings

Of course, that's just my initial thought and is open for discussion.

For reference, here's a screenshot of every app bar state I'm aware of with their current order (every button enabled):
#### Main Menu
MainMenu
#### Subreddit
Subreddit
#### Front Page, Popular, All, Multireddit, Subreddit Search, User Posts, maybe others
Frontpage
#### Comments
Comment
#### Main Menu and Subreddit (two-pane)
MenuAndSub
#### Main Menu and Front Page or equivalent (two-pane)
MenuAndFront
#### Subreddit and Comments (two-pane)
SubAndComment
#### Front Page or equivalent and Comments (two-pane)
FrontAndComment

QuantumBadger:
Thanks! I'll try and set aside some time to review all this soon.

One quick comment regarding `SHOW_AS_ACTION_IF_ROOM` though: I feel like the current behaviour will be highly confusing for users, as the icons won't be shown even if there is plenty of space.

https://stackoverflow.com/a/9039808

I think I'd sooner either remove the option for this, or find a way around it somehow. Comments in the above link suggest overriding the `BaseMenuPresenter` class. From what I can see the logic is contained in [`ActionMenuPresenter.java`](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/widget/ActionMenuPresenter.java#629), and maybe it's possible to just copy the entire source for that class into RR and make the necessary modifications.

Oddly I can't see any reference to the above "2-5 icon" logic in that class, which maybe suggests that it's changed in the latest version of Android, and the Compat libraries we're using have some old logic in.

If going down that route works, we may also be able to avoid hiding the overflow button.

Basically I think this needs some more investigation, but the above might be a useful starting point. I'm happy to help out if you have any questions or run into a dead end!

Cguy7777:
After working on it on and off, I have finally made **Show button if room** work as expected. Trying to override the `Base/ActionMenuPresenter` logic proved difficult, so instead I wrote my own method to figure out how many buttons we can show (`SHOW_AS_ACTION_IF_ROOM` is never actually sent, but converted to `SHOW_AS_ACTION_ALWAYS` or `SHOW_AS_ACTION_NEVER`.)

I also went ahead and created a small method to handle any NullPointerException's from the app bar preference Map, to make the IDE stop warning us about the potential for one.

The only issue I'm aware of remaining is that the options menu isn't invalidated on rotation, which can result in wonky screens where buttons are overflowing or are hidden despite an abundance of space. I'm not exactly sure where a listener to perform this should go, so advice would be appreciated.

There is also still a discussion to be had about any changing of the button order. If keeping the order as close to the current order is desired, I would still recommend putting **Past** after **Refresh** in two-pane mode for consistency, and maybe swapping **Block** and **Subscribe**.

src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
```diff
@@ -82,185 +85,236 @@
 			final Boolean subredditBlocked) {
 
 		final SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(activity);
-		final EnumSet<OptionsMenuItemsPref> optionsMenuItemsPrefs = PrefsUtility.pref_menus_optionsmenu_items(activity, preferences);
+		final Map<Enum<AppbarItemsPref>, Integer> appbarItemsPrefs = PrefsUtility.pref_menus_appbar_items(activity, preferences);
```
QuantumBadger: Looks like this should be an EnumMap :)

src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
```diff
@@ -377,104 +445,152 @@ public boolean onMenuItemClick(final MenuItem item) {
...
 						});
+
+				subscribe.setShowAsAction(showAsAction);
+				subscribe.setIcon(R.drawable.star_off_dark);
+
 				break;
 
 			case UNSUBSCRIBE:
-				menu.add(activity.getString(R.string.options_unsubscribe))
+				final MenuItem unsubscribe = menu.add(activity.getString(R.string.options_unsubscribe))
```
QuantumBadger: Just a minor note for next time: when declaring variables inside a switch statement, I usually put braces around each case, to stop the scope of the variables "leaking" out, and being accidentally used in a future case block. E.g:

```java
switch(var) {
    case 1: {
        int a = 5;
        break;
    }
    case 2: {
        // "a" is not accessible from here
    }
}
```

src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
```diff
@@ -865,14 +867,52 @@ public static boolean network_tor(final Context context, final SharedPreferences
 		return result;
 	}
 
-	public static EnumSet<OptionsMenuUtility.OptionsMenuItemsPref> pref_menus_optionsmenu_items(final Context context, final SharedPreferences sharedPreferences) {
+	private static class AppbarItemInfo {
+		final Enum<OptionsMenuUtility.AppbarItemsPref> itemPref;
```
QuantumBadger: I don't think this needs to be wrapped in `Enum<>`, you can just write the name of the enum :)

QuantumBadger: (same for all usages elsewhere, e.g. line 874)

src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
```diff
@@ -865,14 +867,52 @@ public static boolean network_tor(final Context context, final SharedPreferences
...
+		AppbarItemInfo(final Enum<OptionsMenuUtility.AppbarItemsPref> itemPref, final int stringRes, final int defaultValue) {
+			this.itemPref = itemPref;
+			this.stringRes = stringRes;
+			this.defaultValue = defaultValue;
+		}
+	}
 
-		final EnumSet<OptionsMenuUtility.OptionsMenuItemsPref> result = EnumSet.noneOf(OptionsMenuUtility.OptionsMenuItemsPref.class);
-		for(String s : strings) result.add(OptionsMenuUtility.OptionsMenuItemsPref.valueOf(General.asciiUppercase(s)));
+	public static Map<Enum<OptionsMenuUtility.AppbarItemsPref>, Integer> pref_menus_appbar_items(final Context context, final SharedPreferences sharedPreferences) {
```
QuantumBadger: As before, this can be `EnumMap<OptionsMenuUtility.AppbarItemsPref, Integer>` for improved performance/space usage.

src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
```diff
@@ -865,14 +867,52 @@ public static boolean network_tor(final Context context, final SharedPreferences
...
+
+		Map<Enum<OptionsMenuUtility.AppbarItemsPref>, Integer> appbarItemsPrefs = new HashMap<>();
+
+		for(AppbarItemInfo item : appbarItemsInfo) {
+			try {
+				appbarItemsPrefs.put(item.itemPref, Integer.parseInt(getString(item.stringRes,
+						Integer.toString(item.defaultValue),
+						context,
+						sharedPreferences)));
+			} catch(Throwable e) {
```
QuantumBadger, line 909: Catching `Throwable` isn't ideal -- is this just to catch exceptions thrown by `Integer.parseInt`, and also maybe null pointer exceptions? If so, it should be possible to specify those directly:

```java
} catch(final NumberFormatException | NullPointerException e) {
```

src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
```diff
@@ -865,14 +867,52 @@ public static boolean network_tor(final Context context, final SharedPreferences
...
+				appbarItemsPrefs.put(item.itemPref, item.defaultValue);
+			}
+		}
+
+		appbarItemsPrefs = Collections.unmodifiableMap(appbarItemsPrefs);
```
QuantumBadger: Still possible if you change `HashMap` to `EnumMap`, although the return type of the method would be `Map<OptionsMenuUtility.AppbarItemsPref, Integer>`.

src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
```diff
@@ -602,6 +779,25 @@ public boolean onMenuItemClick(MenuItem menuItem) {
...
 	}
 
+	public static int handleShowAsActionIfRoom(final int showAsAction) {
+
+		if(showAsAction == MenuItem.SHOW_AS_ACTION_IF_ROOM) return MenuItem.SHOW_AS_ACTION_ALWAYS;
+
+		return showAsAction;
+	}
+
+	// Avoids IDE warnings about null pointers
```
QuantumBadger, line 789: A common way to do this is with `Objects.requireNonNull()`, but this is fine :)

src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
```diff
@@ -82,185 +89,285 @@
...
+		if(subredditsVisible && getOrThrow(appbarItemsPrefs, AppbarItemsPref.SETTINGS) == DO_NOT_SHOW) {
+			add(activity, menu, Option.SETTINGS, MenuItem.SHOW_AS_ACTION_NEVER,false);
+		} else {
+			add(activity, menu, Option.SETTINGS, getOrThrow(appbarItemsPrefs, AppbarItemsPref.SETTINGS),false);
+		}
+
+		add(activity, menu, Option.CLOSE_ALL, getOrThrow(appbarItemsPrefs, AppbarItemsPref.CLOSE_ALL), false);
+
+		//Will be called from CommentListingFragment (after potentially adding the Reply button) if comments are shown
+		if(!commentsVisible) pruneMenu(activity, menu, appbarItemsPrefs, !subredditsVisible);
```
QuantumBadger, line 245: It's probably worth calling it from here anyway, even if it's called a second time at some later point. To handle rotation changes, `pruneMenu()` will need to be callable multiple times, and "reset" everything to the correct state each time, and if so calling it twice here isn't an issue.

src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
Cguy7777: Thanks for the tip. :)

src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
Cguy7777: Thanks. I put some braces in just to be safe.

src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
Cguy7777: Thank you for the tip, I actually realized this myself earlier and removed the wrappings.

src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
Cguy7777: I tried to do this, but I couldn't find a way to turn the `Map` `unmodifiableMap()` returns back into an `EnumMap`. I could've just used it as a `Map` in OptionsMenuUtility, but it seems like that would defy the performance-purpose of using an `EnumMap`.

Since we never use any method that could modify the map in there, I think it's probably fine to simply omit this, although if you have a suggestion I'd love to hear it.

Cguy7777:
I've implemented the changes as requested (for the most part - see [my comment](https://github.com/QuantumBadger/RedReader/pull/761#discussion_r467811349) above). Thank you for pointing me in the right direction with the orientation-change issue, it was very easy to implement once I knew where to put it.

QuantumBadger:
Thank you! I've updated the changelog.

However I did notice that the action bar is still missing in the "Action bar items" preferences activity. Do you know if there's a way to add this, so it's consistent with the rest of the preferences?

PR #762, opened 2020-07-23
keep screen on during video

ebaldridge:


QuantumBadger:
Thanks, but doesn't this keep the screen on whenever any webview is displayed on the screen?

I already put in a fix for this for the internal video player (rather than the internal browser): https://github.com/QuantumBadger/RedReader/commit/44cae6b521bb69cce441605a6f78d8199aa11342

PR #763, opened 2020-07-23
Show pin, block, and subscribe in overflow menu on r/random

Cguy7777:
While working on #761, I discovered that **Pin**, **Block**, and **Subscribe** do not appear in the overflow menu (or as buttons, with the proposed new functionality) when loading a subreddit randomly. This PR fixes this.


QuantumBadger:
Nice work, thanks! Changelog updated: https://github.com/QuantumBadger/RedReader/commit/a70874f23ef553676cf5ab0bd745b63da736ed5f

PR #764, opened 2020-07-26
Show pin, block, and subscribe in overflow menu on r/random in two-pane mode

Cguy7777:
Implements #763 in two-pane mode.

As I have been reminded, always remember to test in every mode. :P


QuantumBadger:
Thanks!

PR #766, opened 2020-07-28
Allow comments button to be hidden

Cguy7777:
Thought it might be nice to be able to hide the Comments button in post listings, for users who never or rarely look at the comments, or access them by flinging. This also adds the option to **View Comments** from the post context menu.


QuantumBadger:
Thanks, this looks good!

I think it would be best not to have the new "View Comments" menu item enabled by default though, as the post context menu is already pretty cluttered :)

Cguy7777:
Good point, most people probably won't access the comments that way. I've removed it as a default.

QuantumBadger:
Thanks!

The new code style check is failing above because there are a few lines longer than 100 chars :) It says there are thousands of warnings, however I've just fixed all of these so you may need to merge from master again.

QuantumBadger:
Btw it's possible to run the above checks locally before pushing using:

    ./gradlew pmd lint checkstyle

Cguy7777:
Thanks, I've fixed it right up for you. :-)

QuantumBadger:
Thank you! Merged, and changelog updated.

PR #778, opened 2020-08-18
Allow videos to be zoomed into

Cguy7777:
After switching to a 3:2 phone, I really started to notice apps that don't let you zoom into videos. With the increasing number of ultrawide screens, I figured the ability to do so would only become more useful over time, so I implemented it here. I also added a preference to zoom in by default.


QuantumBadger:
Thanks! I've tried this out, and the buttons look a bit weird and squashed on devices with narrow screens.

Maybe it would be best to:

* Make the video controls bar take up the full width of the screen
* Hide the floating toolbar (with the info and mute buttons) when the controls are visible, so they aren't visible behind the grey background

Screenshot_1598461312

Cguy7777:
I went ahead and implemented the requested changes. Sorry about the mess.

Cguy7777:
Erm... actually, this is a bit excessive with the changes. I'll open a new request.

PR #779, opened 2020-08-18
Allow video to be scrubbed with single taps

Cguy7777:
I thought it was a bit odd how you can only scrub a video with a drag motion, and not by simply tapping on the desired spot on the seek bar. This allows that.


QuantumBadger:
Thanks! Changelog updated: https://github.com/QuantumBadger/RedReader/commit/38126bd075a4d81d8af9e0fdd65106b3c399d2e8

PR #780, opened 2020-08-19
Fix context menu strings

Cguy7777:
While looking through the settings, I realized I made a mistake in #741 when I decided to refer to several context menus as "Long-press" menus, because it is possible to change the ways that those menus are activated. I've seen the term "Action Menu" used in the behaviour preferences to refer to them, and I think that would be preferable.


QuantumBadger:
Looks fine, thanks!

PR #784, opened 2020-08-27
Allow videos to be zoomed into

Cguy7777:


QuantumBadger:
Nice work, thanks! Changelog updated.

PR #790, opened 2020-09-30
Display Post ID in post and comment Properties dialogs

ertyghu:


QuantumBadger:
Thanks for the PR! However outside of development stuff I don't think there's much of a use case for this, so I'd err on the side of not adding clutter to the properties menus.

PR #796, opened 2020-10-24
Support the Enter key on physical keyboards

windsurfer:
I'm using an Fxtec Pro1 and the enter key wasn't working for this field.


QuantumBadger:
Thanks! Changelog updated here: https://github.com/QuantumBadger/RedReader/commit/3dc75985cadc2219732a4f268e617be6f32df9cf

PR #798, opened 2020-10-28
Add optional submenu system for quick account switching

### LINKED ISSUE #755: Simpler/faster account switching

Cguy7777:
This adds an option to switch accounts through a submenu when selecting **Accounts** rather than opening the full accounts dialog. If you choose to have **Accounts** as a button on your appbar, it now takes only two taps to switch accounts. :-)

Whether or not this should be the default for new users may be a question worth considering (as it is now, it's **not** the default).

Closes #755.


QuantumBadger:
Very nice work, thanks! I'd like to make this the default behaviour.

Couple of minor suggestions:

* I'd replace the "Account Manager" item with two items, "Add Account..." and "Manage...", to make things clearer for new users
* Ideally the "Add account..." and "Manage..." items should have icons to the left ("plus sign" and "cog" respectively) to help visually distinguish them from the account names

Cguy7777:
I've updated the branch in line with the recent changes, and enabled quick account switching by default. Right now, it would be the default for both current and new users; typically we try to maintain the current behavior, but I don't imagine too many users would want to revert. But I could make it disabled by default for current users, if so desired.

I'm not sure that splitting the account manager into two options would be a better experience. To me, it seems a bit wasteful of screen real estate to have two buttons for relatively rare actions (adding/deleting accounts); for someone with many accounts and/or a small phone, it could be the difference between having all accounts fit on-screen and having to scroll the menu. If the current setup would be confusing for new users, perhaps a different string or a mention in the first-run dialog could solve that.

I agree that having an icon by the special button(s) is a good idea regardless: I've put a cog icon onto the current account manager button to show how that looks.

QuantumBadger:
Thanks! I've updated the changelog and fixed the preference being off by default here: https://github.com/QuantumBadger/RedReader/commit/9fa62ea4594cdf7df5ae11aae851138e2b90ade6

There's a minor bug which is that the settings cog icon doesn't change colour based on the theme (the dark icon is used whether the theme is light or dark) -- would you mind having a look at this please @Cguy7777?

PR #800, opened 2020-10-31
Make age display customizable

Cguy7777:
Taking inspiration from @QuantumBadger's comment [here](https://old.reddit.com/r/RedReader/comments/jhturs/request_make_the_posts_and_comments_age_shorter/gabd40k/), I decided to take a stab at this. With this, you can now choose the number of time units that will be displayed. You can show only one, show every unit from years to milliseconds, or anywhere in-between.

In addition, I added a preference to allow users to switch between three modes for comment ages:
- Absolute: As it is now: if it says "1 day ago", then the comment was posted one day ago.
- Relative to post: All comments' ages are relative to the post they're on. "1 day ago" means it was posted one day after the original post.
- Relative to parent comment: All comments' ages are relative to the comment they're replying to (top-level comments are still relative to the post).

Some of the code might not be perfect: in particular, I wonder if there is a better way to get access to the ages of parent-comments and posts then this currently uses in RedditCommentView.

Also, I'm not sure what the ideal string would be for comment replies: right now it's "replying %s later", which works okay but is pretty long. It should probably be self-explanatory enough, to prevent confusion.

In any case, I look forward to receiving feedback on this and making any necessary improvements. :-)

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/adapters/SessionListAdapter.java
+++ b/src/main/java/org/quantumbadger/redreader/adapters/SessionListAdapter.java
@@ -112,7 +112,13 @@ public class SessionListAdapter extends HeaderRecyclerAdapter<RecyclerView.ViewH
 		final BetterSSB name = new BetterSSB();
 
 		if(RRTime.utcCurrentTimeMillis() - session.timestamp < 1000 * 120) {
-			name.append(RRTime.formatDurationFrom(context, session.timestamp), 0);
+			name.append(
+					RRTime.formatDurationFrom(
+							context,
+							session.timestamp,
+							R.string.time_ago,
+							2),
+					0);
 		} else {
 			name.append(RRTime.formatDateTime(session.timestamp, context), 0);
 		}
--- a/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
+++ b/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
@@ -812,6 +812,20 @@ public final class PrefsUtility {
 				sharedPreferences)));
 	}
 
+	public static int appearance_inbox_age_units(
+			final Context context,
+			final SharedPreferences sharedPreferences) {
+		try {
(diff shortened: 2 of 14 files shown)


QuantumBadger:
Good stuff, thanks for looking at this!

Just a quick comment, I think people will have trouble understanding what "Post/comment maximum age units" means. I'd call the preference "Post age precision", and provide the following three options:

* "One unit (x hours)"
* "Two units (x hours, x mins)"
* "Three units (x hours, x mins, x secs)"

Similarly, I'd include examples in the "Comment age mode" preference:

* 'Time since comment ("x mins ago")'
* 'Relative to post ("after x mins")'
* 'Relative to parent comment ("replied x mins later")'

Otherwise this looks good to me!

Cguy7777:
I've updated the branch and implemented the requested changes. Thank you for the string suggestions, I thought that the titles of the preferences weren't great, but I wasn't able to think of something better at the time.

One thing I forgot to mention earlier: Currently, inbox messages are hard-coded at two units (just messages; comments in your inbox use the **Comment age precision** preference). I considered adding a separate preference for this, but it seemed excessive to me to have a preference just for that. But maybe it should have one?

QuantumBadger:
Great, thanks!

> Currently, inbox messages are hard-coded at two units (just messages; comments in your inbox use the Comment age precision preference). I considered adding a separate preference for this, but it seemed excessive to me to have a preference just for that. But maybe it should have one?

Up to you, if you'd like to add a preference for this, under a new "Inbox" header in the appearance settings, then that would be fine by me. Ideally the inbox preference would be used for both comments and messages in the inbox.

Cguy7777:
After a little thought, it seems this would be good to have, so I've implemented it. The new **Inbox entry age precision** preference controls all items in the inbox (both messages and comments).

QuantumBadger:
Nice work, thanks! I've updated the changelog, and also removed the quotes from the settings text as I think it looks cleaner.

PR #803, opened 2020-11-10
Combine Nested if statements

bd0n4lds:
Found these two if statements which can be consolidated by separating their conditions.


QuantumBadger:
Thanks!

PR #804, opened 2020-11-13
Add dialog for clearing the cache

### LINKED ISSUE #235: Cache clear buttons in cache prefs screen
Add buttons in the cache preferences screen that allows you to clear image cache or post list/comments cache, showing on each how much space is being taken up by each cache

I.e.,

```
+----------------------------+
| Clear Image Cache (511 MB) |
+----------------------------+

+---------------------------+
| Clear Post Cache (2.1 MB) |
+---------------------------+
```

### LINKED ISSUE #366: [feature request] Delete images and videos from cache
I use your app a lot and the cache usage in android goes up to 1-1,5 GB. I want to clean it sometimes but I don't because I want to keep the cached posts/listings. An option in the settings in the cache section for this would be great.

Sorry for posting all of these feature requests at once, [I posted most of them on the subreddit](https://www.reddit.com/r/RedReader/comments/4nje7r/feature_requests_4_small_features_i_have_been/) about a month ago but you probably missed that post :)

Cguy7777:
This adds a dialog in the **Cache** preferences that allows you to manually clear the cache at any time, by groupings (Listings, Thumbnails, Images, Flags). The dialog also shows you how much space each one of those is currently using (except Flags).

Closes #235, closes #366.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/cache/CacheDbManager.java
+++ b/src/main/java/org/quantumbadger/redreader/cache/CacheDbManager.java
@@ -307,4 +307,31 @@ final class CacheDbManager extends SQLiteOpenHelper {
 		final SQLiteDatabase db = this.getWritableDatabase();
 		db.execSQL(String.format(Locale.US, "DELETE FROM %s", TABLE));
 	}
+
+	public synchronized HashMap<Long, Integer> getFilesToSize() {
+		final SQLiteDatabase db = this.getWritableDatabase();
+
+		final Cursor cursor = db.query(
+				TABLE,
+				new String[] {FIELD_ID, FIELD_TYPE},
+				null,
+				null,
+				null,
+				null,
+				null,
+				null);
+
+		final HashMap<Long, Integer> filesToCheck = new HashMap<>(32);
+
+		while(cursor.moveToNext()) {
+			final long id = cursor.getLong(0);
+			final int type = cursor.getInt(1);
+
+			filesToCheck.put(id, type);
+		}
+
+		cursor.close();
+
+		return filesToCheck;
+	}
 }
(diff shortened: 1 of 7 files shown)


japanesephundroid:
Thanks for this 👍 and wow its been 4 years since that feature request, 5 for the one before me.

@QuantumBadger when you get the time, we are here to test this.

QuantumBadger:
Looks good! I'll take a closer look at this, but just a couple of initial comments:

* We shouldn't do the filesystem activity (looking up the file sizes) on the UI thread -- if we block the UI thread for more than a few seconds, Android will kill the app. It would be better to do the checks in a new thread, and pop up a progress dialog in the meantime.
* Files may be deleted in the background while we're looping through the list from the database (CacheManager.java:252), so we should make sure `file.length()` doesn't throw and cause the bug report activity to start

Cguy7777:
As requested, I've moved the data usage checks (plus the actual cache pruning) to separate threads. Instead of a **ProgressDialog**, I thought it'd be a better idea to put a **ProgressBar** in the dialog and update the **TextView**'s strings (e.g. from "Listings" to "Listings (x MiB)") when the file checks are done. That way, a user who wants to clear caches regardless of their data usage can do so without waiting on the checks.

To be honest, I'm not very familiar with threads in Android, so I'd recommend taking a close look at that code to be safe.

I looked into preventing issues with `file.length()`, and I think we're fine there: according to [the documentation](https://developer.android.com/reference/java/io/File#length()), `length()` will simply return `0L` if the file doesn't exist.

QuantumBadger:
Thank you! I've updated the changelog and done some minor tidying here: https://github.com/QuantumBadger/RedReader/commit/f970ff913af33636062ac80c8bacb504250e89d3

The threading stuff looked fine, although I've modified it to use `AndroidCommon.runOnUiThread()` rather than creating a new `Handler`.

QuantumBadger:
@japanesephundroid This should be in Alpha 238 if you wanted to take a look :)

japanesephundroid:
I tried it, seems to be working OK.

It was instantaneous so I am guessing that it is working in the background like you said above. I only deleted, images/thumbnails.

My cache was not that big, I will update when it gets very big with any issues.

PR #806, opened 2020-11-16
Show correct icon for account manager depending on theme

Cguy7777:
I forgot to make sure that the icon for the **Manage...** submenu in the quick account switcher added in #798 always matches the current theme. This fixes that.

Thanks for catching this quickly @QuantumBadger.


QuantumBadger:
Thanks for fixing this so quickly!

PR #807, opened 2020-11-16
Show tooltips for subreddit actions menu.

### LINKED ISSUE #751: No tooltips on new subreddit toolbar
When long selecting a button on the new subreddit toolbar, no tooltip, toast, or similar appears to describe what the button does.

Tested on both 1.10 and the current codebase ([latest commit when I tested](https://github.com/QuantumBadger/RedReader/commit/3fe213395cbf8f38b82db9806a84a82fc9884167)) on API 27 and API 16.

LiamMeron:
Added tooltip info to subreddit_header_toolbar.xml information. Closes #751

### DIFF
--- a/src/main/res/layout/subreddit_header_toolbar.xml
+++ b/src/main/res/layout/subreddit_header_toolbar.xml
@@ -36,7 +36,8 @@
 			android:padding="12dp"
 			android:background="?rrThemedButtonSelector"
 			android:contentDescription="@string/action_subscribe_subreddit"
-			android:src="?rrIconStarFilledOff"/>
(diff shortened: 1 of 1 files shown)


LiamMeron:
#751

QuantumBadger:
Thanks for the pull request! Unfortunately the `tooltipText` property was added in Android API 26, but the app supports devices from API 14 onwards.

It does look like it's possible to set this programatically though:

https://stackoverflow.com/a/51936412

Maybe instead of doing this in XML, the above method could be used in Java code to set the tooltip text equal to the `contentDescription`.

QuantumBadger:

PR #809, opened 2020-11-19
Inline images

windsurfer:
Adds the option to see images inline on a post listing if they are cached. Default is off.

The images are loaded from disk and resized on a separate thread so scrolling performance isn't affected during the resize.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/cache/CacheManager.java
+++ b/src/main/java/org/quantumbadger/redreader/cache/CacheManager.java
@@ -18,6 +18,7 @@
 package org.quantumbadger.redreader.cache;
 
 import android.annotation.SuppressLint;
+import android.app.Activity;
 import android.content.Context;
 import android.net.Uri;
 import android.os.Build;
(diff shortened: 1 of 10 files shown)


QuantumBadger:
Thanks for the pull request! Unfortunately I can't accept this in its current form:

* When I toggled the setting, nothing happened and images weren't displayed inline. Looking through the code, it seems like `notifyCached()` is invoked from the comment precache callback rather than the image precache callback :)
* Even after fixing this locally it still didn't work, it turned out this was because I was on a mobile connection and no precaching was taking place. I think a lot of users would want to have inline images on a mobile data connection without downloading/precaching the full resolution image every time, and it would be confusing for this to not work with no explanation.
* The height of the post subtitle section is too large, particularly when the thumbnail size setting is turned up.
* lower section height
* Swiping posts left and right is broken, I'd expect the post title and the image to be included in the swipe.
* Images "pop in" a lot and change the scroll position. It would be good if the view set itself to the right dimensions, and displayed a loading spinner until the image was available.

I'd sooner:

* Use the reddit preview images inline rather than the full resolution image
* Keep the reddit post layout the way it currently is (i.e. not split) and have the inline image below the post
* Disable precaching when using inline images
* Show videos inline as well, although this seems like a separate (more difficult) task!

As I already have ideas about how I'd like to implement this, it's probably best if I code it up myself this weekend. I do appreciate the pull requests though, thank you!

PR #810, opened 2020-11-21
App strings reworked

comradekingu:


src/main/res/values/strings.xml
```diff
@@ -23,7 +23,7 @@
     <!-- First Run Dialog -->
 
     <string name="firstrun_login_title">Log in</string>
-    <string name="firstrun_login_message">You are not currently logged in. You can access the account list by selecting "Accounts" from the menu.</string>
+    <string name="firstrun_login_message">You are not logged in. Access the account list by selecting \"Accounts\" from the menu.</string>
```
QuantumBadger, line 26: I prefer the original here. The wording is deliberately "you can" because the first-run dialog already contains a button to log in, so the message is just talking about a future possibility rather than something the user should do now. Also prefer the first sentence with "currently".

src/main/res/values/strings.xml
```diff
@@ -251,7 +251,7 @@
     <string name="action_report_sure">Are you sure you want to report this?</string>
 
     <string name="action_save_image">Save Image</string>
-    <string name="action_save_image_success">Saved to Pictures folder as</string>
+    <string name="action_save_image_success">Saved to '\Pictures\' folder as</string>
```
QuantumBadger, line 254: Backslash after the quote, I think this is causing the build to fail.

src/main/res/values/strings.xml
```diff
@@ -280,15 +280,15 @@
...
+    <string name="post_captcha_title">CAPTCHA</string>
     <string name="post_captcha_submit_button">Submit</string>
     <string name="post_submit_done">Post submitted.</string>
 
     <!-- Bug Reporting -->
 
     <string name="bug_title">An error has occurred.</string>
-    <string name="bug_message">RedReader has encountered an error.\n\nWould you like to send a bug report to the developer? If so, an email will be filled in for you with all the relevant information, and you can inspect the contents before sending.</string>
-    <string name="bug_button_send">Send bug report by email</string>
+    <string name="bug_message">Would you like to send a bug report to the developer? If so, an e-mail with all the relevant info will be filled in for you, and you can inspect it all before sending.</string>
```
QuantumBadger, line 290: Prefer the original wording

src/main/res/values/strings.xml
```diff
@@ -280,15 +280,15 @@
...
+    <string name="bug_button_send">Send bug report by e-mail</string>
```
QuantumBadger, line 291: I generally prefer "email" to "e-mail"

src/main/res/values/strings.xml
```diff
@@ -338,7 +338,7 @@
     <string name="error_postlist_cache_message">You are viewing a past version of this content, and no more posts from this session are in the cache. Press refresh to download the latest posts, or select a different past version from the menu.</string>
 
     <string name="error_postlist_redditdown_title">Reddit is Down</string>
-    <string name="error_postlist_redditdown_message">Reddit is under heavy load at the moment. Press refresh to download the latest posts, or select a different past version from the menu.</string>
+    <string name="error_postlist_redditdown_message">Reddit is under heavy load at the moment. Press \'Refresh\' to download the latest posts, or select a different past version from the menu.</string>
```
QuantumBadger, line 341: The button isn't labelled "Refresh" (it's an icon) so this could be misleading. Would be better as "the refresh button".

src/main/res/values/strings.xml
```diff
@@ -375,7 +375,7 @@
     <string name="props_url">URL</string>
     <string name="props_num_comments">Number of Comments</string>
     <string name="props_body_markdown">Body (Raw Markdown)</string>
-    <string name="props_self_markdown">Self text (Raw Markdown)</string>
+    <string name="props_self_markdown">Self-text (Raw Markdown)</string>
```
QuantumBadger, line 378: I don't think reddit hyphenate "self text" (applies throughout the diff)

src/main/res/values/strings.xml
```diff
@@ -463,7 +463,7 @@
 
     <!-- 2013-05-18 -->
 
-    <string name="action_fling_disabled">Disabled</string>
+    <string name="action_fling_disabled">Off</string>
```
QuantumBadger, line 466: Is this for inclusivity reasons? Maybe clearer as "No action".

src/main/res/values/strings.xml
```diff
@@ -664,7 +664,7 @@
     <!-- 2015-08-15 -->
     <string name="action_delete">Delete</string>
     <string name="delete_confirm">Are you sure you want to delete this?</string>
-    <string name="delete_success">Successfully deleted. The item will still be visible until you refresh.</string>
+    <string name="delete_success">Deleted. The item will still be visible until you refresh.</string>
```
QuantumBadger, line 667: Prefer original

src/main/res/values/strings.xml
```diff
@@ -680,18 +680,18 @@
     <string name="album_already_first_image">Already at first image in album.</string>
     <string name="album_already_last_image">Already at last image in album.</string>
 
-    <string name="upgrade_v190_login_message">RedReader has upgraded, and now uses oAuth for reddit authentication. If you were signed in to an account before the upgrade, you are now logged out.</string>
+    <string name="upgrade_v190_login_message">This new version of RedReader uses OAuth to log in on Reddit. Any account in use before the upgrade is now logged out.</string>
```
QuantumBadger, line 683: This deliberately mentioned upgrades as it's only shown after upgrading, not on new installs. I'd probably just leave this string alone as I doubt it's been shown to anyone in years :smile:

src/main/res/values/strings.xml
```diff
@@ -749,33 +749,33 @@
 
     <!-- 2016-04-03 -->
     <string name="error_tor_not_installed">Tor will not work unless Orbot is installed. Install now?</string>
-    <string name="error_tor_setting_failed">Failed to set proxy.</string>
+    <string name="error_tor_setting_failed">Could not set proxy.</string>
 
     <!-- 2016-04-10 -->
     <string name="error_parse_imgur_title">Parse Error</string>
-    <string name="error_parse_imgur_message">Imgur sent some data I couldn\'t understand.</string>
+    <string name="error_parse_imgur_message">Imgur sent incomprehensible data.</string>
```
QuantumBadger, line 756: Preferred the original. Imgur might not be at fault here :p

src/main/res/values/strings.xml
```diff
@@ -749,33 +749,33 @@
...
 
     <string name="error_upload_fail_imgur_title">Upload failed</string>
-    <string name="error_upload_fail_imgur_message">The upload was rejected by Imgur for an unknown reason.</string>
+    <string name="error_upload_fail_imgur_message">Rejected by Imgur for unknown reason.</string>
```
QuantumBadger, line 759: Unnecessarily terse, preferred the original

src/main/res/values/strings.xml
```diff
@@ -749,33 +749,33 @@
...
 
     <string name="upload_to_imgur">Upload to Imgur</string>
     <string name="no_file_selected">No file selected</string>
     <string name="button_upload">Upload</string>
     <string name="button_browse">Browse…</string>
 
     <string name="error_file_too_big_title">File too big</string>
-    <string name="error_file_too_big_message">The file size (%1$s) is larger than the upload limit (%2$s).</string>
+    <string name="error_file_too_big_message">Ensure the file size (%1$s) is smaller than the upload limit (%2$s).</string>
```
QuantumBadger, line 767: Preferred original

src/main/res/values/strings.xml
```diff
@@ -828,7 +828,7 @@
 	<string name="pref_behaviour_fling_comment_right_title">Fling right</string>
 
 	<!-- 2016-07-17 -->
-	<string name="comment_header_search_thread_title">You are currently viewing search results for a comment thread.</string>
+	<string name="comment_header_search_thread_title">Viewing search results from a comment thread.</string>
```
QuantumBadger, line 831: The original "for" made more sense

src/main/res/values/strings.xml
```diff
@@ -870,7 +870,7 @@
 	<string name="pref_cache_location_title">Cache storage location</string>
 
 	<string name="error_cache_dir_does_not_exist_title">Cache directory missing</string>
-	<string name="error_cache_dir_does_not_exist_message">The cache directory could not be accessed. Please choose a different location in the cache settings.</string>
+	<string name="error_cache_dir_does_not_exist_message">Could not access cache directory. Please choose a different location in the cache settings.</string>
```
QuantumBadger, line 873: Overly terse, I prefer not to omit "the" in error messages/etc (except in toasts, where terseness is good)

src/main/res/values/strings.xml
```diff
@@ -1022,7 +1022,7 @@
 	<!-- 2018-01-16 -->
 	<string name="lang_fi" translatable="false">Suomi</string>
 	<string name="lang_my" translatable="false">မြန်မာဘာသာ</string>
-	<string name="lang_nb_rNO" translatable="false">Bokmål</string>
+	<string name="lang_nb_rNO" translatable="false">Norwegian Bokmål</string>
```
QuantumBadger, line 1025: Should this be "norsk"? (the string should be the Norwegian word for the language rather than the English word)

src/main/res/values/strings.xml
```diff
@@ -1162,7 +1162,7 @@
 	<string name="search_results_subreddit_only">Search on /r/%s</string>
 
 	<!-- 2020-02-17 -->
-	<string name="pref_network_tor_summary">When using an external browser or Custom Tabs, links will not be opened using Tor (unless your browser is configured to use Tor)</string>
+	<string name="pref_network_tor_summary">When using an external browser or custom tabs, links will not be opened using Tor (unless your web-browser is configured to use Tor)</string>
```
QuantumBadger, line 1165: Custom Tabs is deliberately capitalised to match Google's usage:

https://developers.google.com/web/android/custom-tabs

Also I've never seen "web browser" hyphenated :p

src/main/res/values/strings.xml
```diff
@@ -1181,13 +1181,13 @@
 
 	<!-- 2020-02-20 -->
 	<string name="upgrade_message">Updated to version %s</string>
-	<string name="error_toast_failed_open_external_browser">Failed to open url \"%s\" in external browser</string>
-	<string name="error_toast_no_email_apps">No email apps installed!</string>
-	<string name="bug_chooser_title">Email bug report</string>
+	<string name="error_toast_failed_open_external_browser">Could not open the URL "%s" in any web-browser</string>
```
QuantumBadger, line 1184: Preferred original, "any" could include the internal browser

src/main/res/values/strings.xml
```diff
@@ -1181,13 +1181,13 @@
...
+	<string name="error_toast_no_email_apps">No e-mail apps installed.</string>
+	<string name="bug_chooser_title">E-mail bug report</string>
 
 	<!-- 2020-06-08 -->
-	<string name="subscription_successful">Successfully subscribed to %s</string>
-	<string name="unsubscription_successful">Successfully unsubscribed from %s</string>
+	<string name="subscription_successful">Subscribed to %s</string>
+	<string name="unsubscription_successful">Unsubscribed from %s</string>
```
QuantumBadger, line 1190: Preferred original

src/main/res/values/strings.xml
```diff
@@ -1267,13 +1267,13 @@
 
 	<!-- 2020-08-18 -->
 	<string name="pref_behaviour_video_zoom_default_key" translatable="false">pref_behaviour_video_zoom_default</string>
-	<string name="pref_behaviour_video_zoom_default_title">Zoom into videos by default</string>
+	<string name="pref_behaviour_video_zoom_default_title">Zoom in on videos by default</string>
```
QuantumBadger, line 1270: Preferred original

src/main/res/values/strings.xml
```diff
@@ -1267,13 +1267,13 @@
...
 
 	<!-- 2020-08-23 -->
 	<string name="image_gallery_removed_title">Gallery removed</string>
-	<string name="image_gallery_removed_message">This gallery has been removed and is no longer accessible.</string>
+	<string name="image_gallery_removed_message">This removed gallery is no longer accessible.</string>
```
QuantumBadger, line 1274: Original made more sense

src/main/res/values/strings.xml
```diff
@@ -1283,8 +1283,8 @@
 	<string name="image_gallery_no_image_present_message">The specified image could not be found. It may have been deleted.</string>
 
 	<!-- 2020-09-04 -->
-	<string name="reddit_login_browser_popup_line_1">To login, you will now be taken to Reddit in a web browser.</string>
-	<string name="reddit_login_browser_popup_line_2">When the login is complete, you should be redirected back to RedReader. If the login fails, you can try a different browser using the checkbox below.</string>
+	<string name="reddit_login_browser_popup_line_1">To log in, you will now be taken to Reddit in a web-browser.</string>
```
QuantumBadger, line 1286: No need for the hyphen

src/main/res/values/strings.xml
```diff
@@ -1283,8 +1283,8 @@
...
+	<string name="reddit_login_browser_popup_line_2">After logging in, you should be redirected back to RedReader. Try a different browser if you can\'t log in by using the checkbox below.</string>
```
QuantumBadger, line 1287: Preferred the original (although "After logging in" is fine)

src/main/res/values/strings.xml
```diff
@@ -1303,17 +1303,17 @@
 	</plurals>
 
 	<!-- 2020-09-19 -->
-	<string name="action_save_image_success_no_path">Image saved successfully</string>
+	<string name="action_save_image_success_no_path">Image saved</string>
```
QuantumBadger, line 1306: Preferred original

src/main/res/values/strings.xml
```diff
@@ -1303,17 +1303,17 @@
...
 
 	<!-- 2020-10-20 -->
 	<string name="pref_behaviour_imagevideo_tap_close_key" translatable="false">pref_behaviour_imagevideo_tap_close</string>
 	<string name="pref_behaviour_imagevideo_tap_close_title">Close images/videos on tap</string>
-	<string name="pref_behaviour_imagevideo_tap_close_summary">Does not affect videos if playback controls are enabled</string>
+	<string name="pref_behaviour_imagevideo_tap_close_summary">Does not affect videos if playback controls are on</string>
```
QuantumBadger, line 1311: Preferred original

src/main/res/values/strings.xml
```diff
@@ -1303,17 +1303,17 @@
...
 
 	<!-- 2020-11-13 -->
 	<string name="press_back_again">Press back again</string>
 	<string name="pref_behaviour_postlist_back_again_key" translatable="false">pref_behaviour_back_again</string>
-	<string name="pref_behaviour_postlist_back_again">Press back twice to exit post list</string>
+	<string name="pref_behaviour_postlist_back_again">Press \'Back\' twice to exit the post list</string>
```
QuantumBadger, line 1316: Similar to other comment, the button isn't labelled "Back", so I'd prefer this as "Press back twice to exit the post list"

QuantumBadger (changes requested):
Thanks for the pull request! There are some good changes in here, but in a lot of cases I prefer the original string, in which case I've left a comment.

It looks like the build is failing, I think this is because a backslash was put after a quote rather than before it, I've left a comment on the offending line :)

src/main/res/values/strings.xml
comradekingu, line 26:
```suggestion
    <string name="firstrun_login_message">You are not logged in. The account list can always be found under \"Accounts\" from the menu.</string>
```

src/main/res/values/strings.xml
comradekingu, line 254:
```suggestion
    <string name="action_save_image_success">Saved to \'Pictures\' folder as</string>
```

src/main/res/values/strings.xml
comradekingu: It is already under "an error has occurred", so that part is redundant.
Then it self-references the program, which is bad practice. "contents" doesn't go beyond "relevant info", so that is superfluous. "Information" isn't different from "Info".
The sum of changes, which keep the wording the exact same for other parts, manages to be a lot shorter, which helps comprehension, and induces less stress.

src/main/res/values/strings.xml
comradekingu: The word "successfully" is at best redundant. If not, it suggests this often fails in a way that unsubscribes, but not successfully so. It is also longer, and software doesn't do this. This is debug style info in the UI, which isn't good.

src/main/res/values/strings.xml
comradekingu: You don't zoom into something. Certainly not something flat.

src/main/res/values/strings.xml
comradekingu: Why? When did it get removed?

src/main/res/values/strings.xml
comradekingu: Consistent with pre-fetch

src/main/res/values/strings.xml
comradekingu: Original is a needless arbitration to what the user is doing. It also isn't different from the user doing it, as when done, the login is also complete. Fixed to "web-browser"

comradekingu, line 1287:
```suggestion
	<string name="reddit_login_browser_popup_line_2">After logging in, you should be redirected back to RedReader. Try a different web-browser if you can\'t log in by using the checkbox below.</string>
```

src/main/res/values/strings.xml
comradekingu: Is an image ever saved unsuccessfully?

src/main/res/values/strings.xml
comradekingu: Are they ever turned on and disabled?
Is there even options to do both?
If not, it is either off, or on.

comradekingu, line 1311:
```suggestion
	<string name="pref_behaviour_imagevideo_tap_close_summary">Only affects videos without playback controls.</string>
```

src/main/res/values/strings.xml
comradekingu: Is it "←"?

src/main/res/values/strings.xml
comradekingu: Produces errors in translation though. The dash isn't optional in many other languages. Easier to omit when not there than to include in error.

src/main/res/values/strings.xml
comradekingu: It would be better as the actual icon, in UTF-8.

src/main/res/values/strings.xml
comradekingu: Pretty sure I checked, will find out.

src/main/res/values/strings.xml
comradekingu: No, disabled people are still disabled. This presumably is either "On" or "Off".(?)

src/main/res/values/strings.xml
comradekingu: If it is in there, it should be consistent. I can comment it out if you want?

src/main/res/values/strings.xml
comradekingu: Who is "I", is the bigger question I am left with.
Arguably comprehension is on part of RedReader.

src/main/res/values/strings.xml
comradekingu: The user knows it is trying to upload something. There is no need to suger-coat this non-descriptive error in non-descriptive language padding.

src/main/res/values/strings.xml
comradekingu: All the others are invocations to do something. Why then should this be just information that the user is meant to figure out?

src/main/res/values/strings.xml
comradekingu: Why? It isn't on part of, it is from a source.
"Viewing comment thread results" isn't that good, and neither is "in".(?)

src/main/res/values/strings.xml
comradekingu: It is consistent. It is more terse only in the interest of brevity. The original is longer and less terse.

src/main/res/values/strings.xml
comradekingu, line 1025:
```suggestion
	<string name="lang_nb_rNO" translatable="false">Norsk bokmål</string>
```

comradekingu: That is just something Wikipedia is doing.

src/main/res/values/strings.xml
comradekingu: Yes, that presumably also failed?

src/main/res/values/strings.xml
comradekingu: From the brilliant minds that wrote "If you are getting started with Custom Tabs, _checkout_ the Implementation Guide".
Should it be "external file-browser" or "external web-browser" btw?

src/main/res/values/strings.xml
QuantumBadger: The sentence "((an email with X) will be filled in)" is semantically different to the original "(an email (will be filled in with X))". Also the second version uses the word "all" twice which sounds clunky.

Just seems like needless rewording of something that was marginally clearer before :p Let's just leave it as it was originally.

src/main/res/values/strings.xml
QuantumBadger: The other translations can use "e-mail" if that's more appropriate for those particular languages, for English we should stick with "email" though.

The word "Mail" in those three examples is used in the sense of "message from another user", so yes it is internal reddit mail, but the English translation should stay as "mail".

src/main/res/values/strings.xml
QuantumBadger: I think that could end up more confusing, particularly if the font is small :p

src/main/res/values/strings.xml
QuantumBadger: This is an option under Behaviour -> Post Actions -> Fling left/right. "Disabled" is one of a series of options such as "Upvote", "Downvote", etc.

"Off" seems slightly weird in that list, "No action" could work but lacks the implication that flinging in that direction is prevented. So leaving this as "Disabled" is probably best.

src/main/res/values/strings.xml
QuantumBadger: It's still used by the code, but will only appear in the very specific case that someone upgrades from v1.8.7.8 (from several years ago) to a newer version.

Every second spent worrying about this string is a second wasted, so let's just leave it as it was :smile:

src/main/res/values/strings.xml
QuantumBadger: The original was fun and quirky, whereas the new one is terse, frigid, and arguably accusatory. Let's just leave it as it was :smile:

src/main/res/values/strings.xml
QuantumBadger: I prefer padding out error messages if it makes them sound less cold/terse/ungrammatical

src/main/res/values/strings.xml
QuantumBadger: The new wording sounds slightly odd due to the numbers in the string, in the same way that "please ensure 5 is less than 3" doesn't make sense. I think it's generally just clearer as it was.

src/main/res/values/strings.xml
QuantumBadger: On re-reading I agree with you, let's make this "from" then. However I do prefer the "You are currently" in the original, it's clearer and friendlier :)

src/main/res/values/strings.xml
QuantumBadger: Longer, more grammatical sentences are often easier to read/comprehend that ones which have all grammatical niceties stripped out. In this particular case I don't mind too much either way.

src/main/res/values/strings.xml
QuantumBadger: Not sure what you mean about Wikipedia -- basically all I'm saying is that these language strings should be in the target language, rather than English. For example, `lang_de` is "Deutsch" rather than "German".

src/main/res/values/strings.xml
QuantumBadger: Haha :smile: Even so, let's keep it capitalised as it makes it clearer that it's referring to the Android feature. In fact I had someone message me a few weeks ago asking what it meant, so maybe best as "Android Custom Tabs" for maximum clarity.

It should be "web browser" (no hyphen please)

src/main/res/values/strings.xml
comradekingu: True, but it breaks down to the same resulting e-mail, containing the same info.

comradekingu, line 290:
```suggestion
    <string name="bug_message">Send the bug report to the developer? If so, an e-mail with relevant info will be filled in that you can inspect before sending.</string>
```

src/main/res/values/strings.xml
QuantumBadger: The string is used in cases where the external browser was explicitly requested (e.g. due to settings) and the internal browser wasn't tried.

src/main/res/values/strings.xml
comradekingu: @QuantumBadger It isn't appropriate. I am saying people misunderstand it in translation, to mean "arrived e-mail" and "has e-mail address".

src/main/res/values/strings.xml
QuantumBadger: "External browser" is also consistent with preferences/etc, and may encompass things that aren't strictly web browsers (e.g. the YouTube app) if they are set up to handle certain links

src/main/res/values/strings.xml
QuantumBadger: Okay, let's go with the new version

src/main/res/values/strings.xml
QuantumBadger: After giving this some thought we should entirely change this to "Expand videos to fill screen" :smile:

src/main/res/values/strings.xml
QuantumBadger: It's deeply unidiomatic and relegates the most essential information ("gallery has been removed") to a lesser position, implying that it's something the user should already been aware of. Let's keep the original please.

src/main/res/values/strings.xml
QuantumBadger: "Prefetch" and "precache" should also not be hyphenated

src/main/res/values/strings.xml
QuantumBadger: The new version is a garden path sentence -- it initially reads as "(Try a different browser) if you can't (log in by using the checkbox below). Please revert!

QuantumBadger: Also web browser was originally correct, it's not hyphenated in English!

src/main/res/values/strings.xml
comradekingu: Could say the "🗘" icon on the top left (?), or somesuch

src/main/res/values/strings.xml
comradekingu: The software (hopefully) isn't sentient.

src/main/res/values/strings.xml
comradekingu: Well if there isn't a denomination it doesn't make sense in either case. Should I add a "MB"?

src/main/res/values/strings.xml
QuantumBadger: Let's keep it as it was, I don't think this is an improvement.

src/main/res/values/strings.xml
QuantumBadger: We shouldn't modify strings just to provide context for translators, when the context is evident in the UI for the user. It would be better to leave a comment on Weblate for future translators.

src/main/res/values/strings.xml
comradekingu: Flipping the verb, subject and objective around like this isn't a nicety. Why the cache directory has to be a marked noun I don't know. I can add the "the" again if you want.

src/main/res/values/strings.xml
QuantumBadger: That would be less accessible for screenreaders, let's leave it as it was

src/main/res/values/strings.xml
comradekingu: "E-mail" is also the proper form, consistent with other such words. You may disagree, but people that don't exist. "Email" just recently came into fashion. "E-mail" will be there till it all becomes "mail" and regular mail is either that or "snailmail".

src/main/res/values/strings.xml
QuantumBadger: In general I don't subscribe to the idea that cutting out words from status messages is a goal unto itself, that always makes things more readable. Redundancy is a natural part of language, and often improves clarity.

I don't really mind one way or another in this particular case, I think the original was fine as we should just leave it as-is.

src/main/res/values/strings.xml
QuantumBadger: The denomination is added programmatically in `ImgurUploadActivity`

src/main/res/values/strings.xml
QuantumBadger: Okay thanks, let's go with the new version but add the "the" again

src/main/res/values/strings.xml
QuantumBadger: I've gone ahead and removed two of these strings as they were unnecessary and mostly debug-related when the app was first written: https://github.com/QuantumBadger/RedReader/commit/f70d8aa7ca9e3e90b1df174eec184992930c6357

The third one ("Mod Mail") is the term Reddit themselves use, although it looks like they actually use one word "Modmail", so feel free to change it to that: https://old.reddit.com/r/modguide/comments/dlc0v9/how_to_use_modmail/

I agree "email" and "e-mail" are both valid, but since (1) "email" is in fashion, and (2), RedReader already uses "email", let's keep it as it was.

src/main/res/values/strings.xml
comradekingu: @QuantumBadger It is 53 chars shorter, and _still_ verbose … ("Would you like to - a " can be omitted.) How is that not a whole lot better?

The original is awful. There is an "if so" conditional, hinging on the first sentence, and the second doesn't end before piling on "relevant" and "info", referencing that in "all"; which no questions answered …
The "you" is invoked for no good reason, and it is posed as a question that just creates doubt.
The whole thing pivots on a question that may or may not have something to be inspected, that the user should or should not already be knowledgeable about prior to answering.

**An error has occurred.**
Would you like to send a bug report to the developer? If so, an e-mail with all the relevant info will be filled in for you, and you can inspect it all before sending.
[Send bug report by email] [Don\'t send bug report]

**Reportable error**
Inspect the prefilled e-mail before sending it to the developer.
[Send] [Cancel]

The level of entropy here is "it" refers to both the e-mail and error. "Reportable" ties them together. There is a flow that doesn't treat the user as a dumpsite for thoughtless text.

There is an error, it is reportable. Sending it in happens via e-mail, and the e-mail needs to be inspected before it is sent. When it is sent, the recipient will be the developer.
Instead of the apologetic style, "reportable" is good. Fail graciously.

QuantumBadger: @comradekingu We've spent way too long debating this. This appears for very rare, literally one-in-every-million-sessions errors, so verbosity is not an issue. Let's not spend any more time on it please!

src/main/res/values/strings.xml
QuantumBadger: It's the Android system back button, which is normally a left-pointing triangle, but the actual look is system-dependent. This was fine as is.

src/main/res/values/strings.xml
comradekingu: Solid point.

src/main/res/values/strings.xml
QuantumBadger: This doesn't preserve the meaning, it's specifically referring to the video playback controls preference, which can be enabled or disabled.

src/main/res/values/strings.xml
comradekingu: > cutting out words from status messages is a goal unto itself, that always makes things more readable.

I agree, not always though. It can't be both redundant and more clear. While fewer words aren't always better, all redundant words can and should be cut.

src/main/res/values/strings.xml
comradekingu: Disabled seems like it is beyond the control of the user to turn it on or off. Less complexity is good.

src/main/res/values/strings.xml
comradekingu: To me it seems "pre-cache" is the verb, whereas "precache" is the noun. I much prefer the action, and both return hits.

src/main/res/values/strings.xml
comradekingu: Great. I care and you don't, so that is you opting to put time into keeping it worse.
The original is worse for reasons like opting to use "bug report", which is not as common, and hard to translate in some languages. Keep it simple and solid and translators will put a lot more effort into trying to keep up with the quality.

src/main/res/values/strings.xml
comradekingu: There is also "reddit" and the double "you". "signed in" isn't used elsewhere. It still says upgrade. Not twice now, which helps the user figure out what an upgrade even is. That is better.

src/main/res/values/strings.xml
```diff
@@ -749,33 +749,33 @@
...
     <string name="image_selected_summary">Image selected: %1$dx%2$d, %3$s</string>
 
 	<!-- 2016-06-10 -->
 	<string name="submit_post_specify_subreddit">Specify subreddit</string>
-	<string name="submit_post_title_empty">Title cannot be empty</string>
-	<string name="submit_post_url_empty">URL cannot be empty</string>
-	<string name="submit_post_self_text_hint">Self Text</string>
+	<string name="submit_post_title_empty">Add a title first</string>
+	<string name="submit_post_url_empty">Add a URL first</string>
+	<string name="submit_post_self_text_hint">Self-post text</string>
```
comradekingu, line 778: This is the type of garbage English has instead of proper compound word rules. I dropped the optional hyphen in web-browser, but this will almost certainly fail in translation.

comradekingu, line 778:
```suggestion
	<string name="submit_post_self_text_hint">Text for self post</string>
```

src/main/res/values/strings.xml
```diff
@@ -1303,17 +1303,17 @@
 	</plurals>
 
 	<!-- 2020-09-19 -->
-	<string name="action_save_image_success_no_path">Image saved successfully</string>
+	<string name="action_save_image_success_no_path">Image saved</string>
 
 	<!-- 2020-10-20 -->
 	<string name="pref_behaviour_imagevideo_tap_close_key" translatable="false">pref_behaviour_imagevideo_tap_close</string>
 	<string name="pref_behaviour_imagevideo_tap_close_title">Close images/videos on tap</string>
```
comradekingu, line 1310:
```suggestion
	<string name="pref_behaviour_imagevideo_tap_close_title">Tap to close images/videos</string>
```

QuantumBadger:
I appreciate how much you care about this, but I think the best thing to do is for me to manually merge in the specific strings I think it's beneficial to change.

Going forwards, I'm not planning to accept future pull requests that re-word existing English strings (from anyone), because whether one wording is better than another is highly subjective, and the time spent debating it could be spent on improving the app in other ways.

Thank you for submitting this though -- as I say I do appreciate you trying to make the app better.

comradekingu:
@QuantumBadger I care about the string-base for Hosted Weblate, and this app doesn't pass for mustard, nor does it follow any standards. I don't make any changes I can't explain, and that aren't also consistent. There is plenty of changes that improve on minute things too, but you seem to be under the impression that changes are the enemy of preserving whatever quality you think is there.
The README says "reddit" line one, but I can't help you there either.

QuantumBadger:
> The README says "reddit" line one, but I can't help you there either.

When RedReader was first developed over seven years ago, the Reddit brand guidelines said that the word "reddit" must always be in lowercase. That has since changed, and so I've updated the README (and the strings).

> There isn't any context or consistency, and the other translations are all over the place.

There's some context available in the key of the string (e.g. `pref_appearance_fontscale_post_subtitles_title` tells you that it's the name of a preference), but I agree the situation could be better.

Translation context shouldn't be added by changing the strings themselves. I'm happy to go through on Weblate and add comments to each string, but this won't be finished for a while as there are nearly 900 strings. If you have other suggestions for how to make translators' lives easier (that don't involve modifications to the English text) then please let me know.

Anyway let's stop the discussion now over the string changes please. The thing that's using up my time here is the back-and-forth debate over each minute change -- the issue isn't that I don't care or don't want to make changes, it's that I thought a lot of the changes were detrimental or incorrect, and that's my call to make as the project maintainer. There were some changes in there which were better, which I've now taken manually, so no need to discuss further.

comradekingu:
@QuantumBadger 
Somehow your project isn't in good shape. Despite this, my way of doing is subject to the most resistance possible. Yet, you want some of it done, and discourage any further improvement.

You can remove whatever you think is superfluous as "context" from the strings. How this is bad when adding arbitrary baggage under guise of being clearer is beyond me. It already failed in translation, and it doesn't get easier from there for any of the users.
Contextualizing the strings that are problematic doesn't fuzzify the current erroneous translations, but it is a good start. That is 121 strings, but I marked quite a few for a later pass, as there are many improvements to be found. You adamantly don't care.
My best guess is other translators will come along to patch up the unfinished translations instead of being able to do it all at once, or get some overview of things.

If a lot of the changes are detrimental, that is news to me. That a lot of them are incorrect, is incorrect. Your time is yours, your project is yours, and the strings in a "reader" aren't good.

I wish it was all better, successfully.

comradekingu:
@QuantumBadger Over a year on, the app has only has coverage in 9 languages, and it is all over the place even there.
Absolute word-salad is a problem even though it can be deciphered by those already in the know.
Translating this is extremely difficult, and doing a good job is harder still.

Can you comment on the strings you find to be irrelevant/strange/or worse, and I can explain just those?
There is a method to it all, and no strings are changed just for the sake of it.

QuantumBadger:
@comradekingu This was discussed extensively at the time, and I don't think it would be productive to re-litigate it now.

If there are difficulties translating the app, the solution isn't to modify the English strings, but rather to add context to the strings using Weblate. Anyone should be able to do this by leaving comments on Weblate.

PR #820, opened 2020-12-06
Shows reddit post comments when title clicked

mgurga:
It seems a bit redundant that clicking the title of a post shows the image, so I included the title LinearLayout inside the comment LinearLayout. This effectively makes the title part of the comment button while keeping the same visuals.


QuantumBadger:
Thanks for this, but it looks like it wouldn't allow people to read web posts, or to view images/videos if inline previews are turned off?

I'm not looking to change the default behaviour of this, but I'd consider accepting a PR which adds a preference (off by default), and which works by modifying the click listeners rather than making everything part of the comments button (which feels a bit hacky).

PR #823, opened 2020-12-11
Don't auto-collapse your comments or user-profile comments

### LINKED ISSUE #314: Don't apply min. comment hiding when viewing comments on a user's profile
And probably shouldn't apply to the user's own comments, if they drop below the threshold.

Cguy7777:
This prevents comments from being automatically collapsed due to the **Minimum comment score** if

- The comment is by you (the current user), or
- You are viewing comments on a user profile

Closes #314.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/InboxListingActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/InboxListingActivity.java
@@ -290,7 +290,9 @@ public final class InboxListingActivity extends BaseActivity {
 												parsedComment,
 												null,
 												-100000,
+												null,
 												false,
+												true,
 												true);
 
 										itemHandler.sendMessage(General.handlerMessage(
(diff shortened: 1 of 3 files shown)


QuantumBadger:
Thanks! Just one minor issue, the username comparison should be case insensitive. It's probably best to do this using `StringUtils.asciiLowercase()` to avoid locale issues (I'd change the `currentUserName` param to `currentUserNameLowercase` to avoid converting it every time).

Cguy7777:
No problem! I've switched it to use the canonical username (lowercase and trimmed).

Cguy7777:
@QuantumBadger It looks like the unit tests failed because it couldn't download Gradle to run them. The last commit was quite minor, this is probably good.

QuantumBadger:
Perfect, thanks! Changelog updated here: https://github.com/QuantumBadger/RedReader/commit/19db5f8530d5bb404a1d76f19bcf2cf0d5f2095a

PR #838, opened 2021-01-25
Add preference for minimum comment height

Cguy7777:
I often tap the wrong collapsed comment when holding my phone certain ways (in my defence, I have my system UI set to small). As a result, I was inspired to add an accessibility preference to set a minimum comment height, so people who experience motor-control issues, or who are simply prone to fat-fingering, will have an easier time.

While this is mainly targeted at making collapsed comments easier to hit, this is a minimum for the whole comment; at higher levels, this will also start to affect non-collapsed comments (if they're short enough).


QuantumBadger:
Looks good, thanks! Merging this in.

QuantumBadger:
Changelog updated.

PR #840, opened 2021-01-30
Add more post/comment fling actions

Cguy7777:
I noticed that several actions for posts and comments available through the action menu aren't available as quickly-accessible fling actions. This adds many of those actions as options for that.

I included every action I felt could plausibly be used widely and regularly; actions that are for subreddit management (subscribing, pinning, or blocking subs), only applicable to user posts/comments (editing, deleting), or only applicable to self-text (since you can't see the self-text without opening it) were not added.


QuantumBadger:
Nice work, thanks! Merging this in.

QuantumBadger:
Changelog updated: https://github.com/QuantumBadger/RedReader/commit/d6434d4cb154ca7717fbd78df37831faf68f5080

PR #841, opened 2021-02-02
Do not close loading images when tapped

### LINKED ISSUE #819: Touching the screen while an image is loading causes it to cancel
Tapping or swiping the screen at all while an image is loading has the same effect as tapping the back button. This only occurs on v1.15.

Disabling the new setting to tap to close images has no effect on this. It can be a little annoying to accidentally touch the screen while a large image is loading and have to start over.

Cguy7777:
This prevents loading images from being closed by touching the screen. I've tested this with TalkBack and it doesn't seem to affect the behavior there, although it might be a good idea to double-check.

Closes #819.


QuantumBadger:
Looks fine to me, thank you! Merging this in.

PR #845, opened 2021-02-17
Add preference to customize comment action menu items

Cguy7777:
There are a lot of different items in the Comments **Action Menu**, so it seemed like a good idea to bring this in-line with other context menus in the app by adding the ability to customize it.


QuantumBadger:
Perfect, thanks! I've updated the changelog.

PR #846, opened 2021-02-17
Improve byte-unit display

Cguy7777:
This changes `addUnits()` so that a size of less than 10 KiB/MiB/GiB will show the first decimal place. This will provide more detail on the amount of storage available or in use.

This is particularly useful with something over a gibibyte. Currently, if you see "1 GiB", it's unclear if you're using exactly one gibibyte, nearly two gibibytes, or somewhere in-between.


QuantumBadger:
Looks good, thanks!

PR #852, opened 2021-03-08
Disable certain logging in release builds

### LINKED ISSUE #851: System logging should be checked
@QuantumBadger Currently, a lot of information is written to the system log; some of this information represents a potential data leak to apps with the ability to read that log (see [SC-10](https://developer.android.com/docs/quality-guidelines/core-app-quality#SC-10)).

Potential solutions:
- Use ProGuard to strip logging calls from the release builds.
- Create custom methods for logging: in the debug build these would simply pass arguments to `Log`, and in the release build they would ignore or reduce logging requests. (Alternatively, a preference for logging could be added, instead of relying on build type alone.)

Cguy7777:
This disables all logs mentioned [here](https://github.com/QuantumBadger/RedReader/issues/851#issuecomment-792227147) in release builds. I also corrected a log in `NewMessageChecker` that was set to ERROR instead of INFO.

Closes #851.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/AlbumListingActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/AlbumListingActivity.java
@@ -65,7 +65,9 @@ public class AlbumListingActivity extends BaseActivity {
 			return;
 		}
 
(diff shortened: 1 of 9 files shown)


QuantumBadger:
Looks good, thanks!

PR #858, opened 2021-03-16
Add ability to view sent messages

### LINKED ISSUE #346: Add ability to view sent messages
Similar to inbox, but for sent messages

Cguy7777:
Decided to implement this. I was unable to fully test if Modmail still works correctly since I'm not a moderator of any subreddit, so be sure to check that. I have *not* set this to appear in the **User items** by default, but I think it probably should be there for easy discovery, and I can make it so if desired.

I used email-send by Michael Richins from [Material Design Icons](https://materialdesignicons.com/) for the menu icon.

Closes #346.


QuantumBadger:
Thanks! I've merged this in manually and updated the changelog.

It looks like the sent messages list doesn't currently show who you sent the message to, which would be useful :)

I think at some point it would be good to revamp the way inbox messages work in general in the app, to allow for e.g. viewing conversation threads rather than just individual messages.

PR #859, opened 2021-03-19
Add content descriptions for media buttons

Cguy7777:
I noticed that buttons in both the floating image buttons and the media playback controls were unlabeled for screen readers, and added descriptions for them.


QuantumBadger:
Thank you! Just checking, are the playback controls actually accessible using a screen reader?

Cguy7777:
I'm not aware of a way to open the playback controls with TalkBack running, although you can easily select the floating Image Info and Mute buttons. We should probably look into a way for screen-reader users to access those controls easily.

QuantumBadger:
Thanks! Merged in manually and changelog updated.

PR #863, opened 2021-04-11
Show recipients of sent messages

Cguy7777:
As mentioned in #858.

PS: I think it'd be a good idea to show **Sent Messages** by default for new users, to aid discovery.


QuantumBadger:
Nice work, thanks! Merged in.

> PS: I think it'd be a good idea to show Sent Messages by default for new users, to aid discovery.

Ideally I'd like to avoid adding more items to the main menu, but I agree it would be good to make this more visible by default. Maybe we could add it to the three-dots menu in the inbox?

PR #865, opened 2021-04-15
Implement toggles for thumbnails and previews on spoiler posts

### LINKED ISSUE #864: [FEATURE REQUEST] Add "Show spoiler previews" and "Show spoiler thumbnails" options
Hi,

As the title says, I was hoping to have the option to enable and disable showing of spoiler previews and thumbnails. This feature is already available for NSFW posts, so I don't think it would be too difficult to implement (I've only taken a cursory glance at the code and don't know Java very well, so I don't really know). These options would be found in the Images/Video section of the settings, like it's NSFW counterpart.

Thanks!

FintasticMan:
Closes #864

I don't have any experience writing Java, and haven't tested this. I couldn't tell whether `post.spoiler` was the correct way to tell if a post is a spoiler, so could someone verify that? The comment on line 845 of PostListingFragment.java should be removed once that is done.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
+++ b/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
@@ -121,6 +121,8 @@ public final class PrefsUtility {
 				|| key.equals(context.getString(R.string.pref_images_inline_image_previews_key))
 				|| key.equals(context.getString(
 						R.string.pref_images_inline_image_previews_nsfw_key))
+				|| key.equals(context.getString(
+						R.string.pref_images_inline_image_previews_spoiler_key))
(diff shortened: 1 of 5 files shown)


FintasticMan:
I've tested this using the debug build from the CI, and everything seems to work.
I've also removed the comment about `post.spoiler`, because it works.
closes #864

QuantumBadger:
Thanks for this! I've had a quick look and it seems fine to me. When I get a free moment I'll take a closer look and get this merged in.

FintasticMan:
Awesome, thank you!

QuantumBadger:
Looks like this all works correctly -- I've merged it in, updated the changelog, and rearranged the strings in date order. Thanks again!

PR #875, opened 2021-06-28
Better highlighting of posts/comments

### LINKED ISSUE #62: Highlight moderators, admins in comments thread, posts list

Cguy7777:
This highlights posts by admins and moderators, and highlights posts/comments by special authors when reading out-loud for screen readers. Feel free to comment on my proposed strings for this; in particular, I think "OP" might be a better term than "submitter", but also might be easy to mistake for extra syllables of the author's name?

One thing to note: comments, made by an admin/moderator that submitted the original post, are currently only highlighted as being from the OP. Maybe we should add a visual indicator that they are also an admin/mod? Old Reddit [shows both](https://old.reddit.com/r/announcements/comments/dpqd0z/the_extra_life_charity_award_raise_awareness_for/f5xpwav/).

Closes #62.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditParsedPost.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditParsedPost.java
@@ -255,6 +255,10 @@ public class RedditParsedPost implements RedditThingWithIdAndType {
 		return mSrc.author;
 	}
 
+	public String getDistinguished() {
+		return mSrc.distinguished;
+	}
+
 	public String getRawSelfTextMarkdown() {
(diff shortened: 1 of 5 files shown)


QuantumBadger:
Perfect, thank you! I've just merged this in and updated the changelog. Also updated the parcelling logic on `RedditPost` to handle the new `distinguished` field.

> One thing to note: comments, made by an admin/moderator that submitted the original post, are currently only highlighted as being from the OP. Maybe we should add a visual indicator that they are also an admin/mod? Old Reddit shows both.

Yes, that sounds like a good idea -- the same format as old reddit (i.e. the coloured letters inside square brackets) would be fine.

PR #877, opened 2021-07-05
Wrap preference strings, and don't reserve space for icons

### LINKED ISSUE #750: Preference labels do not wrap

Cguy7777:
This wraps preference strings that are too long to show on one line. It also looked like a good idea to not reserve space for icons, since we aren't using any.

Thank you for migrating to the new AndroidX preferences, which made this possible.

Closes #750.


QuantumBadger:
Thank you! Changelog updated.

PR #878, opened 2021-07-05
Add date sorting for Controversial posts & user comments

### LINKED ISSUE #666: Add date sorting for other sort types
Only "Top" can be sorted by day, month, year, all time. Please make other sort types like controversial, best etc. be able to do the same.

Cguy7777:
With this, you can now sort Controversial by hour/day/week/month/year/all on posts and user comments. I also noticed and fixed a small error in the human path creator in `UserPostListingURL`.

Closes #666. (The poster of this issue wants to be able to sort *all* sort types; Reddit does not provide all sorts with date sorting, so this is not possible, unless we add our own manual sorting logic — quite a task. I think we can close that issue?)


QuantumBadger:
Nice work, thank you!

> Closes #666. (The poster of this issue wants to be able to sort all sort types; Reddit does not provide all sorts with date sorting, so this is not possible, unless we add our own manual sorting logic — quite a task. I think we can close that issue?)

Agreed, we don't want to implement our own manual sorting.

PR #880, opened 2021-07-06
Inform user there are no posts to show when necessary

Cguy7777:
This shows a bit of text when there are no posts to display. Currently, if no posts are to be shown when getting a list of posts, the user simply sees a blank screen. Adding this will give a more polished appearance, and assure users that there is truly nothing to see (rather than leaving them wondering if the app/connection had a glitch).

One thing to note: this does show a message while still downloading posts, if there are no posts to show *yet* due to settings. The message will disappear if posts to display are found. I think this is nice, because it can give the user an impression of why a search is taking awhile (due to the app filtering out many of the posts retrieved).

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/adapters/RedditListingManager.java
+++ b/src/main/java/org/quantumbadger/redreader/adapters/RedditListingManager.java
@@ -18,12 +18,8 @@
 package org.quantumbadger.redreader.adapters;
 
 import android.content.Context;
-import androidx.annotation.StringRes;
 import androidx.recyclerview.widget.LinearLayoutManager;
-import android.view.LayoutInflater;
 import android.view.View;
-import android.widget.TextView;
-import org.quantumbadger.redreader.R;
 import org.quantumbadger.redreader.common.General;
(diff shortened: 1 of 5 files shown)


QuantumBadger:
Thanks, this looks good!

I've given it a go, and it seems like the text can appear before the loading spinner has disappeared. I assume this is because the app is trying to load a second page of posts, but nothing was shown from the first page.

Is it possible to delay this until after the spinner is gone please? See below for a video of the issue.

Also, just a minor cosmetic thing, the text should smoothly transition in if you add `mEmptyItem` to the group only when necessary, rather than having it there from the start and setting the contents to visible.

https://user-images.githubusercontent.com/2998158/125161043-eb4d4680-e177-11eb-9773-f2a1edd137c0.mp4

Cguy7777:
Thanks for the feedback! I took a look at the code again and realized that it is simpler, as well as smoother looking, to add the view in only when needed. I just needed to find a way to make sure that we're done loading any more posts before we add it. I think we should be good here. :-)

QuantumBadger:
Perfect, thanks! I've merged this in and updated the changelog.

PR #885, opened 2021-07-18
Add ability to use new controversial date-sorts as defaults, convert old prefs on upgrade.

Cguy7777:
I realized that in #878, I forgot to add an option to use the new controversial sorts as defaults in the settings, and to make sure they upgrade gracefully if originally set to **Controversial** (without a time period).


QuantumBadger:
Thank you! I've merged this in here: https://github.com/QuantumBadger/RedReader/commit/29713f9ec9042d7d50c63460c6db68b1482d1698

PR #886, opened 2021-07-18
Add date sorting for search sorts

Cguy7777:
Similar to my earlier PR, this adds date sorting for searches. There's no need to mess with any preferences here, since there isn't an option to set a default sort for searches (although maybe that's something we should add?)

I did my best to match the original code; since *every* search sort can be sorted by time period, it's a bit verbose. I could potentially do some reworking, but maybe it's best to leave it as is for ease of understanding, despite the length?

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/OptionsMenuUtility.java
@@ -1114,11 +1114,98 @@ public final class OptionsMenuUtility {
 			sortPosts.getItem().setShowAsAction(handleShowAsActionIfRoom(showAsAction));
 		}
 
-		addSort(activity, sortPosts, R.string.sort_posts_relevance, PostSort.RELEVANCE);
-		addSort(activity, sortPosts, R.string.sort_posts_new, PostSort.NEW);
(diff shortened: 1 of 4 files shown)


QuantumBadger:
Nice work as usual, thanks! I've merged this in and updated the changelog.

Just a minor code quality point, it might be worth considering some static HashMaps in `parseSearch()` rather than the long list of if statements, for example:

```java
private static HashMap<String, PostSort> SEARCH_POST_SORTS_RELEVANCE = new HashMap<>();

static {
    SEARCH_POST_SORTS_RELEVANCE.put("all", RELEVANCE_ALL);
    SEARCH_POST_SORTS_RELEVANCE.put("hour", RELEVANCE_HOUR);
    // ... etc ...
}
```

And then do the lookup using `HashMap.get()`. It's a minor point though, the current code works fine and matches what we're already doing in `PostSort.java` :)

> there isn't an option to set a default sort for searches (although maybe that's something we should add?)

I'd happily accept a pull request for this, so if you'd like to add it please feel free :)

QuantumBadger:
Actually now that I think about it, it would be possible to use reflection to do the whole thing in one step:

```java
PostSort.valueOf((General.toAsciiLowercase(sort) + (t == null ? "" : "_" + General.toAsciiLowercase(t)));
```

It's good to avoid reflection where possible though, so again this is fine :)

PR #889, opened 2021-08-01
Massive cleanup of PrefsUtility

TacoTheDank:
### 1st commit
- Add a gradle wrapper validation action (see [here](https://github.com/gradle/wrapper-validation-action))

### 2nd commit
- Update gradle wrapper to 7.1.1 using git bash

### 3rd commit
- Update Studio to 4.2.2
- Update buildTools to 30.0.3
- Update NDK 21.4 -> 22.1

### 4th commit
I've refactored the PreferenceHelper to use a single sharedPreferences object initialized on app launch.

Benefits:

- No more passing a bunch of unnecessary contexts and sharedPreference objects around.
- No more unnecessary operations where we create a brand new sharedPreference instance any time a key needs to be accessed.
- Note that PreferenceManager.getDefaultSharedPreferences was never deprecated, you just had to update PreferenceManager to its AndroidX version.
- This might actually improve performance a tiny bit.

I intend to expand further on this cleanup (if acceptable) in another PR, as there's a lot more that can be done. I wanted to start out small so you don't get overwhelmed by all the changes or anything.

### 5th commit
- Cleaned up the GitHub workflow so we aren't wasting time doing the same thing over and over for each step before each thing actually runs.


QuantumBadger:
Thanks for this, looks like a lot of hard work! It'll take a while to review but hopefully I'll be able to merge it in soon. It definitely does look cleaner without the context being passed around everywhere.

One thing I did notice is the `SharedPrefsWrapper` locking has been removed in most cases. This was designed to prevent the preferences being read/written in the middle of a backup or restore. I'm not sure how necessary or useful this locking actually was though.

Would you mind fixing the Lint issues before I look through this in detail please? At the moment it's failing the CI builds.

TacoTheDank:
@QuantumBadger Done. I also added a commit that cleans up the CI build (shaves off about a minute). It's better that we're not checking out the repo and installing the NDK for every single step.

QuantumBadger:
Thank you! I've gone ahead and merged this in, but there are few tweaks I'll make in a moment.

QuantumBadger:
All done, thanks again :)

This will be in Alpha 291.

TacoTheDank:
Thanks! :)

PR #890, opened 2021-08-03
Update post header when the toolbar is disabled

Cguy7777:
This fixes a bug I discovered, where the subtitle on an open post isn't updated if the comment-list header toolbar is disabled.

For example, if I have the toolbar disabled and upvote an opened post by using the side-toolbar, the vote count does not turn orange, unless I refresh.


QuantumBadger:
Thanks @Cguy7777! I think some of these PRs need rebasing due to the shared prefs changes which just went in :) I can do the rebasing locally if you like, or alternatively feel free to update the PR.

Cguy7777:
@QuantumBadger I went ahead and updated this PR for you. My other two pull requests right now don't touch preferences, and look like they should be good as is.

QuantumBadger:
Perfect, thank you! (and well spotted on the bug)

I've merged this in and updated the changelog.

PR #892, opened 2021-08-09
Block posts from blocked subreddits on the front page

### LINKED ISSUE #653: Block Subreddit does not work.
Subreddits blocked continue to appear on the front page, even after multiple reloads.

### LINKED ISSUE #891: Block Subreddit doesn't work for FrontPage feed.
/r/MakeMeSuffer is the name of my existence on Reddit, can't block it with RR.

I ran into it while using RR as I usually do, scrolling through FrontPage, oh there's a beautiful image of someone who's popped a zit the size of a quarter and there are gobs of white goop everywhere. So I long pressed on the tab of the listing, and set RR to block that subreddit. 

Unfortunately that really didn't do anything at all, and I don't understand why. Today an image of the inside of a woman's skull, bright red, scalp cut open, on a surgery table kissed my vision. I checked, maybe the cache got reset? Nope, it's still blocked. Why do I still see it? 

And yes, I know I can choose to blur out NSFW flagged content, but I'd rather keep that off and just block MakeMeSuffer as it's the only offender.

Version: 1.17.3, downloaded via F-Droid on OnePlus 7 Pro.

Cguy7777:
Posts from blocked subreddits will now be hidden on the front page, which is nice for using the app while logged out.

I also updated the toasts that appear when (un)blocking a sub to reflect this. I kept close to the original wording, but perhaps the strings are a bit long for toasts?

Closes #653, Closes #891.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
+++ b/src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
@@ -785,7 +785,9 @@ public class PostListingFragment extends RRFragment
 											|| mPostListingURL.asSubredditPostListURL().type
 													== SubredditPostListURL.Type.ALL_SUBTRACTION
 											|| mPostListingURL.asSubredditPostListURL().type
-													== SubredditPostListURL.Type.POPULAR);
+													== SubredditPostListURL.Type.POPULAR
+											|| mPostListingURL.asSubredditPostListURL().type
+													== SubredditPostListURL.Type.FRONTPAGE);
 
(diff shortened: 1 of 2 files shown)


QuantumBadger:
This change makes sense, thank you. Merged!

QuantumBadger:
> I kept close to the original wording, but perhaps the strings are a bit long for toasts?

Might work better as either a dialog or a [snackbar](https://material.io/components/snackbars). Dialog might be best as users can dismiss it once they've finished reading.

See `DialogUtils.showDialog()` :)

PR #893, opened 2021-08-12
Show tooltips on the toolbars and side-toolbar

### LINKED ISSUE #751: No tooltips on new subreddit toolbar

Cguy7777:
This adds tooltips to the toolbars added in v1.10. It also replaces the toasts on the side-toolbar with tooltips, for consistency.

Closes #751.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java
@@ -37,6 +37,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.appcompat.widget.TooltipCompat;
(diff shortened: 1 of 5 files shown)


Cguy7777:
Just noticed that this wasn't able to be auto-merged anymore after changes to master, went ahead and resolved that. :-)

Cguy7777:
@QuantumBadger I'm afraid I'm unable to reproduce the glitch.

[This post](https://stackoverflow.com/questions/11792109/linearlayout-cannot-be-cast-to-android-widget-button) seems to suggest that it might solved by cleaning and rebuilding the project. If you haven't, could you try that and see if it resolves the problem? If not, I can investigate further. Thanks. :-)

Cguy7777:
Just tried it with all the new PRs merged, the tooltips work great. :-)

PR #898, opened 2021-08-31
Search multis, and easier seaching of reddit combinations

### LINKED ISSUE #588: Search multi-reddits
When I'm browsing a multi-reddit, I wish the search function could search only within that multi instead of searching all subreddits.

(Apologies if this functionality is already in the app. I haven't found it.)

Cguy7777:
This adds the ability to search multireddits. It also fixes an issue where, when trying to search a reddit combination (r/a+b+c), the search would switch to "all"; in addition, multi URLs using "u" instead of "user" are accepted now, since Reddit accepts those.

I've done some testing with this and it seems to work fine, but some of the code may be a bit awkward, and it should definitely be double checked. Feel free to discuss and raise any concerns.

Closes #588.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/activities/PostListingActivity.java
+++ b/src/main/java/org/quantumbadger/redreader/activities/PostListingActivity.java
@@ -359,6 +359,7 @@ public class PostListingActivity extends RefreshableActivity
 			final SearchPostListURL url;
 
 			if(controller != null && (controller.isSubreddit()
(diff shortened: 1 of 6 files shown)


QuantumBadger:
Sorry for the delay on reviewing all these PRs @Cguy7777, I'll hopefully have some time within the next few days.

QuantumBadger:
Looks good, thanks! Changelog updated.

PR #899, opened 2021-09-01
Allow status bar to be hidden only when viewing media

### LINKED ISSUE #623: [Feature request] Option to hide Android status bar for videos, images, gifs only
Currently we have the option to hide the Android status bar completely.

Another useful option would be to hide it when viewing videos, images or gifs with the internal viewer.

Cguy7777:
Closes #623.


QuantumBadger:
Perfect, thanks! I've updated the changelog.

PR #900, opened 2021-09-06
Handle non-HTTP URLs using native apps in WebView

### LINKED ISSUE #787: Some links redirect to non http links resulting in errors
Example:
https://www.reddit.com/r/LiverpoolFC/comments/iwj260/player_ratings_survey_chelsea_0_2_liverpool_fc/

Follow the "Link to survey" link at the start of the self text.

It redirects to a non http link and gives the following error:

```
Webpage not available

The webpage at intent://forms.gle/?????????????????#Intent;package=com.google.android.gms;action=com.google.firebase.dynamiclinks.VIEW_DYNAMIC_LINK;scheme=https;S.browser_fallback_url=https://docs.google.com/forms/d/e/???????????????????/viewform%3Fusp%3Dsend_form;end; could not be loaded because:

net::ERR_UNKNOWN_URL_SCHEME
```

There are also many examples of this on the Spotify subreddit:
https://www.reddit.com/r/spotify/

Opening any Spotify playlist link gives a similar error.

Using "View in External Browser" correctly passes the link to the Spotify app (or whatever app is set to open that link).

avik-das:
Typically, these are deep links to native apps, which can be opened
using intents. This change attempts to open the native app, falling back
to letting the WebViewClient's URL-handling logic take over if no app is
installed that accepts that intent URL. This not only works for
`intent://` URLs, but also specific ones like `twitter://`.

To verify this change, find a link to a Tweet (found some on r/nfl),
open it using the WebView, then click the "Switch to the app" button
Twitter provides. Before this change, you'd get
`net::ERR_UNKNOWN_URL_SCHEME` erorr in the WebView. After the change, if
you don't have a Twitter client installed, you'll still get the same
error, but if you do have a client installed, it'll be opened.

Closes #787

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/fragments/WebViewFragment.java
+++ b/src/main/java/org/quantumbadger/redreader/fragments/WebViewFragment.java
@@ -362,7 +362,7 @@ public class WebViewFragment extends Fragment
 				final Intent nativeAppIntent;
 				try {
 					nativeAppIntent = Intent.parseUri(url, Intent.URI_INTENT_SCHEME);
-				} catch (URISyntaxException e) {
+				} catch (final URISyntaxException e) {
 					return false;
 				}
 
@@ -389,7 +389,7 @@ public class WebViewFragment extends Fragment
 				try {
 					startActivity(nativeAppIntent);
 					return true;
-				} catch (ActivityNotFoundException err) {
+				} catch (final ActivityNotFoundException e) {
 					return false;
 				}
 			}
--- a/src/main/java/org/quantumbadger/redreader/fragments/WebViewFragment.java
+++ b/src/main/java/org/quantumbadger/redreader/fragments/WebViewFragment.java
@@ -304,6 +304,16 @@ public class WebViewFragment extends Fragment
 					if(RedditURLParser.parse(Uri.parse(url)) != null) {
 						LinkHandler.onLinkClicked(mActivity, url, false);
 					} else {
+						if (!url.startsWith("http:") && !url.startsWith("https:")) {
+							Intent nativeAppIntent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+							try {
+								startActivity(nativeAppIntent);
+								return true;
+							} catch (ActivityNotFoundException err) {
+								// fall through and open the URL as though it's a regular HTTP URL
+							}
+						}
+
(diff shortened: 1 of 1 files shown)


avik-das:
Hi! I recently switched to RedReader because I've been interested in an open source reddit client. Thanks so much for this project, it's quite amazing. Part of why I wanted an open source client was so I could contribute back :)

This particular issue was a pain point for me, and based #787 it was a pain point for others as well. Please let me know if there's any process I need to follow to make accepting this change easier for you.

QuantumBadger:
Thanks for this, looks good! There's a minor PMD check failing, you can run this locally using:

    ./gradlew pmd

As mentioned on #787 I'm a bit concerned about the security implications of allowing `intent://` URLs -- maybe for now these could be blacklisted and we can make a decision at a later point about whether or not to allow them.

avik-das:
Just saw the CI failure, I'll fix it! And totally, I read through #787 in more detail, and I understand your concern. I didn't realize launched intents would use RedReader's permissions, which definitely is a major security vulnerability vector.

I'll see if I can find an example of an `intent://` URL (the issue has an example potentially), and if so, I'll see if I can parse out the fallback URL. Otherwise, I'll just not handle that scheme.

avik-das:
Ok, made the change. Unfortunately, when testing, I found an issue with the Twitter "open in app" link. For some reason, the page that redirects to the `twitter://` URL also links to a `market://` URL to allow installing the native app. The weird part is the second URL is opened _even when_ the Twitter client starts up correctly! So the behavior is that the Twitter app opens, and then the Play Store opens immediately after that.

I'm putting up the code changes in case anyone can see what the issue is while I debug, but I'm not sure this should be merged now. (The test case with the Google Form link in #787 works fine, opening up the fallback URL as expected.)

I added the latest changes as new commits. I'm happy to squash and rebase if you'd prefer a cleaner history. Also, I left a detailed comment in the code because that helps me capture the assumptions behind the code. But if that's not the style of this codebase, I can remove it! Just let me know.

avik-das:
Oops, this time PMD succeeded, but checkstyle failed! I'll fix it.

QuantumBadger:
Nice work, thank you! I've merged this in and updated the changelog. It should be in the next alpha, and also v1.18.

PR #901, opened 2021-09-07
Reply post action menu item

### LINKED ISSUE #896: [Feature Request] Post reply in Post menu
I thought I requested this in the past but can't find the 'issue'..

For some reason, there's no 'Reply' option in the Post action menu.   This has thrown me off for years.  Anytime I want to reply to a post I go to the post menu, and there's no reply option.
I have to keep reminding myself to go into the main menu instead.  Especially because, you click the comment menu to reply to a comment, so it's different behaviours for different comments.

As menus are customizable and people and tick and untick whichever items they want, can we please at least get the option to show the reply menu in the post menu? even if it has to be unticked by default.

Cguy7777:
This adds a Reply option to the post action menu. I think it's a good idea to include this as a default option (which I've done here), for consistency with the comment action menu.

Closes #896.

PS: I accidentally used tabs instead of spaces in part of arrays.xml when I was working on this, and didn't notice until looking over this PR on GitHub, hence the awkward commit history. Maybe it'd be a good idea to switch to all tabs or all spaces in xml files, at least for the most commonly changed ones in `res/values` and `res/xml`?

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java
@@ -335,6 +335,13 @@ public final class RedditPreparedPost implements RedditChangeDataManager.Listene
 						R.string.action_report,
 						Action.REPORT));
 			}
+
+			if(itemPref.contains(Action.REPLY) && !post.isArchived) {
+				menu.add(new RPVMenuItem(
+						activity,
+						R.string.action_reply,
+						Action.REPLY));
+			}
 		}
 
 		if(itemPref.contains(Action.EXTERNAL)) {
--- a/src/main/res/values/arrays.xml
+++ b/src/main/res/values/arrays.xml
@@ -443,6 +443,7 @@
         <item>@string/action_hide</item>
         <item>@string/action_delete</item>
         <item>@string/action_report</item>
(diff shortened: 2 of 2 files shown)


QuantumBadger:
Thanks, looks good!

Ideally it would be good to have this on by default for long-clicking the grey post header in the comments section, but off by default for the menu in the post list. Not sure if there's a clean way to do that though.

Cguy7777:
I don't think there's a particularly good way to have it only show by default in the comments. It probably wouldn't be hard to code in, but we'd either have to disable it in the post listing regardless of user settings, or add a separate preference just for that, which seems a bit excessive to me.

I'd be curious to see if anyone ends up using it frequently from the post listing. You _could_ look at just the post (or even just the title ;-)) and comment without seeing what anyone else has said.

On the other hand, we could simply hide it by default in both locations quite easily.

-----

As for the tabs in XML files, I could go through them and make a PR for that once the current requests are merged in. (I did want to do one or two things involving sorts once #903 is merged or discussed, so perhaps after that?)

QuantumBadger:
Thanks, I've merged this in and updated the changelog.

> As for the tabs in XML files, I could go through them and make a PR for that once the current requests are merged in. (I did want to do one or two things involving sorts once #903 is merged or discussed, so perhaps after that?)

Yep sounds good :) I've merged all the PRs now so this should be possible.

PR #902, opened 2021-09-07
Prevent replying to archived posts and comments

Cguy7777:
While working on #901, I noticed that RedReader doesn't stop you from trying to reply to an archived post or comment. This fixes that, in a way similar to how attempted votes are handled.


QuantumBadger:
Thanks! Looks like this works for the comment header toolbar and probably the action menu, but it doesn't have any effect on the "Reply" option in the three-dots menu in the comment section :)

Cguy7777:
Good catch, I added the check to there as well.

PR #904, opened 2021-09-15
Preserve the letter case of search queries

Cguy7777:
This stops searches from dialog boxes from being forced to lowercase. This is a minor annoyance, since boolean operators [must be uppercase](https://reddit.com/wiki/search) to function, and keeping the case has no ill effects.


QuantumBadger:
Thanks, I'm not sure why this was forced to lowercase in the first place! Merged in.

PR #907, opened 2021-09-22
Don't set over18 cookie or parameter on SFW searches

Cguy7777:
This avoids setting the over18 cookie and parameter for searches when NSFW content is disabled. This can save a bit of bandwidth and loading time, since we're not downloading posts that will be filtered out anyway.

The difference is usually small, but can be very noticeable when making a search that has a lot of NSFW results.


QuantumBadger:
Perfect, thanks! Merged in, and changelog updated.

PR #908, opened 2021-09-22
Add optional indicator (†) for comments voted controversial

Cguy7777:
This adds the ability to show a dagger (or speak a notice) on comments that are voted controversial (lots of upvotes and downvotes), like the website does when enabled in the user's settings.


in-plaintext:
Where can this setting be enabled in the app? I toggled it on in the Reddit settings, if that's necessary:

https://old.reddit.com/prefs/

> [ x ] show a dagger (†) on comments voted controversial (a controversial comment is one that's been both upvoted and downvoted significantly)

The feature should be made available to all Redreader users (the 'anonymous' accounts; if not already).

PR #916, opened 2021-10-14
Add Reply button to post action menu for upgrading users

Cguy7777:
As discussed in #901, this adds the Reply button to the post action menu for upgrading users, making it appear by default for them as well as new installs.


QuantumBadger:
Looks good, thanks!

QuantumBadger:
Changelog updated.

PR #917, opened 2021-10-14
Preserve self-text collapse state on state change

Cguy7777:
This preserves the status of the self-text on self posts (visible or collapsed) across state changes.

Currently, its state is lost, so if a state change occurs (by entering multi-window, (un)folding a folding phone, etc.) the self-text reverts to being opened.


Cguy7777:
On second thought, maybe this isn't the best approach. Perhaps it would be better to store the self-text state in the same way that we store the collapse-state of comments, for consistency? That would preserve the self-text state, even if you leave the post and come back.

QuantumBadger:
Thanks, this looks good. I'm happy to merge in as-is, however storing the collapse state alongside the comments sounds like a good idea too. If you'd like to do a separate pull request implementing that, please feel free :)

QuantumBadger:
Changelog updated.

PR #918, opened 2021-10-20
Show read user posts, even if "Hide read posts" is enabled

### LINKED ISSUE #915: [Bug] "Hide read posts" option hides Starred/Saved posts that have been read
As title, and this is unexpected behavior. Starred posts should always be visible on the `Saved Posts` screen, regardless of their read/unread status.

Cguy7777:
This prevents read posts on any User Post Listing URL (upvoted/downvoted/submitted/saved/hidden posts) from being hidden if **Hide read posts** is enabled.

While I'm certain we don't want to hide the user's own posts from them, I'm not 100% on whether other users' posts should be hidden. I think it's probably more in line with the user's expectations to not hide either, but I could add a check so that *only* the logged-in user's posts are exempt from the hiding.

Closes #915.


QuantumBadger:
Thanks! Merged, and changelog updated.

PR #919, opened 2021-10-21
Prevent replying to locked posts and comments

Cguy7777:
This prevents you from trying to reply to a locked post or comment, and hides the **Reply** option in action menus if possible. Moderators should still be able to reply to locked items on subreddits they moderate.

This resolves part of #303. (This PR doesn't add any indication that something is locked until you try to reply to it. That'd be something nice to add as well.)

PS: I noticed that in **RedditComment**, `score_hidden` is never parceled. Is this intentional?


QuantumBadger:
Sorry for the delay in merging this @Cguy7777 -- now that v1.18 is released I should get around to this soon.

QuantumBadger:
Thank you! I've merged this in, updated the changelog, and also fixed the parcelling issue you mentioned with `RedditComment` :)

Issue #513, opened 2017-06-25
Apply suggested sort for comments


aravindperi:
Some posts have a suggested sort. For example, match threads usually have a New sort suggestion. But RedReader ignores this suggestion and uses the default sort.

Can you add this option? Thanks!

QuantumBadger:
Thanks for the suggestion!

If this gets added, I'd prefer if it was either:

* Toggled using a preference, and off by default
* Or, manually enabled by the user each time, by putting a notification at the top of each comment thread with a suggested sort. E.g. "Subreddit moderators suggest that this thread should be sorted by 'new'. Tap here to change the sort."

Unfortunately I've noticed this feature being used by some mods to bury opinions they don't like (e.g. /r/bitcoin mods sorting block size discussions by "controversial"), but I acknowledge that it has legitimate uses too.

fifty-50:
I prefer the first implementation, rather than manually doing it each time.

clinchergt:
>Toggled using a preference, and off by default

This is already an option on reddit:

![image](https://user-images.githubusercontent.com/1283149/27999469-bc1aedd6-64da-11e7-996c-1018599ea3af.png)

It'd be a bit redundant to add one to the app, don't you think? Unless it uses the API to change that setting.

If no one is working on this feature, I might tackle it soon®

QuantumBadger:
> It'd be a bit redundant to add one to the app, don't you think?

I think the default option on reddit is to enable suggested sorts, whereas in the app I'd sooner it were disabled by default.

There's also the problem of:

* When to fetch the preference -- either we'd need to fetch it on every single comment page load, or we'd need to remove the app's "default comment sort" preference and rely on the site's one for that too (thereby letting the site decide the sort for us).
* Allowing the user to set the preference locally (since some users only use the app). Again this would require fetching the value of the preference every time the settings screen loaded, which isn't really feasible.

So if we do go down the preference route it ought to be local.

> If no one is working on this feature, I might tackle it soon

Sounds good, thanks.

Even with the preference it would be good to have some sort of UI for this, to notify users that a suggested sort is active (and if the preference is disabled, to say that one is available).

PR #922, opened 2021-10-30
Highlight suggested sort in sort menu

Cguy7777:
This shows which sort is the suggested comment sort in the sort menu, when applicable.

I initially started on this with the intention of also making it possible to automatically apply the suggested sort if desired, but realized there are some thorny questions and implementation details to deal with (some discussed in #513). However, simply being able to see which sort is suggested is still useful, so I'll put this forward as is.


QuantumBadger:

> However, simply being able to see which sort is suggested is still useful, so I'll put this forward as is.

Yeah, I think this is good. In the future it might be good to add a message at the top of the comments section saying something like "the suggested sort is X. Would you like to switch to this?"

codeofdusk:
@Cguy7777 The suggested sort should also be indicated in the accessibility description, perhaps in round brackets. For instance, "best (suggested)".

PR #925, opened 2021-11-25
Make videos loop more smoothly

Cguy7777:
I've noticed that videos show a loading icon in the center for a split second when looping. Using the ExoPlayer's method to enable the looping instead seems to fix this.


Cguy7777:
Looks like the CI timed out for some reason, but all of the Gradle checks passed on my machine.

QuantumBadger:
Thank you! Merged in and changelog updated.

PR #941, opened 2022-01-15
Allow side-by-side development build flavor

onemahon:
I'm not sure if you have a preferred way of doing local development without overwriting production builds, but I just threw this together when I was messing around locally. Absolutely feel free to just ignore this if you want!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,6 +71,12 @@ android {
 		proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
 	}
 
+	buildTypes {
+		debug {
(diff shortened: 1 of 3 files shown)


QuantumBadger:
Thanks, I didn't know this was possible!

My only concern is that it creates two identical apps in the app list, both called "RedReader" with the same icon. Do you know if it's possible to also change the name of the launcher entry?

onemahon:
@QuantumBadger yeah, good call! I think we can set it up to have a different name and perhaps a different color app icon or something. I can check into setting that up in a few days!

QuantumBadger:

PR #948, opened 2022-01-26
Add option to show upvote percentage in post subtitles

### LINKED ISSUE #914: upvote percentage
option to display upvote percentage beside upvote count

Cguy7777:
Closes #914.


Cguy7777:
I looked into this, and the only simple way I can see to do it would be to create three strings - one for the bold part, one for everything before that, and one for after. That'd be a little ugly, but it'd probably be worth it to have better language support?

PR #966, opened 2022-03-25
Add access to subreddit Action Menu from post listings

### LINKED ISSUE #727: Add button to "Copy subreddit URL" when viewing a subreddit?
Right now, when you visit a subreddit, you can't copy the URL by long-pressing the black bar on the top of the scroll view. Neither can you open the toolbar's 3-dot menu to copy the URL. Can you add a way to copy subreddit URL?

Right now you can copy a subreddit URL from the home screen, just not the subreddit itself.

Cguy7777:
This allows the header on a subreddit (the area at the top of a post listing that shows the subreddit name and subscriber count or url) to be long pressed to show the subreddit action menu, which is currently only accessible by long pressing a subreddit on the main menu.

While moving the action menu code into its own public static method, I noticed some calls to `setPinnedSubreddits()` and `setBlockedSubreddits()` in `onSubredditActionMenuItemSelected()` that seemed to be unnecessary, and I removed them so that I could make that function static. Everything appears to work correctly without them, but it may be worth double checking that.

Closes #727.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/adapters/MainMenuListingManager.java
+++ b/src/main/java/org/quantumbadger/redreader/adapters/MainMenuListingManager.java
@@ -689,114 +689,7 @@ public class MainMenuListingManager {
 		};
 
 		final View.OnLongClickListener longClickListener = view -> {
-
-			final EnumSet<SubredditAction> itemPref
-					= PrefsUtility.pref_menus_subreddit_context_items();
-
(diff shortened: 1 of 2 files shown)


QuantumBadger:
Perfect, thank you! Agreed that the `setPinnedSubreddits()` and `setBlockedSubreddits()` calls look unnecessary.

PR #967, opened 2022-03-25
Replace default subs r/announcements and r/blog with the new r/reddit

Cguy7777:
Reddit has recently decided to archive r/announcements, r/blog, and r/changelog, and post all new official Reddit news at r/reddit. With that in mind, we should probably replace those old subreddits with the new one in the default list.


QuantumBadger:
Thanks! Merged in.

PR #969, opened 2022-03-25
Make comment level announcement more concise

codeofdusk:
Thanks so much for 461e638! This PR makes the comment level announcement string a bit more speech friendly: dynamic content should be placed as early as possible so users can easily interrupt (for instance, scrolling past comments quickly of a particular level without having to listen to "comment reply" first).


codeofdusk:
Cc @QuantumBadger.

QuantumBadger:
Perfect, thank you! Looks like the linter has a false positive, I'll suppress this and then merge.

QuantumBadger:
Merged in and changelog updated, thanks again :)

PR #970, opened 2022-03-26
Make a few accessibility messages more concise without loss of content

codeofdusk:
This PR makes minimal changes to some accessibility strings while increasing concision.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,6 +87,7 @@ android {
 		// Translations are crowd-sourced
 		disable 'MissingTranslation'
 		disable 'MissingQuantity'
+		disable 'ExtraTranslation'
 
 		// Temporarily disabled
 		disable 'ImpliedQuantity'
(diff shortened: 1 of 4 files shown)


codeofdusk:
CC @QuantumBadger.

Note that this PR is separate from #971, where my intention is to have more drastic fixes behind a user preference. This PR focuses on minimal changes that shouldn't require more Reddit knowledge or otherwise change the messages in a major way.

QuantumBadger:
Thanks, I've had a look and this looks good!

One minor problem is that if a comment has one upvote, this will read out "one up votes" in the plural. There's an example of how to do plurals on line 1289 of strings.xml:

https://github.com/QuantumBadger/RedReader/blob/1a63fbc5d7203dc17400fdca3ed554d7f0afeddd/src/main/res/values/strings.xml#L1289

Otherwise this looks good, thank you!

codeofdusk:
@QuantumBadger I've changed the definition of `accessibility_subtitle_points_withperiod` in commit b07113a, but the build is failing as it can't find the string definition. Something up with my syntax?

QuantumBadger:
Thank you! Looks like `R.string` needs to be changed to `R.plurals` in the two places where it's used.

QuantumBadger:
Sorry, I think there's some extra steps involved in changing the plural usage. I'll investigate and push a fix :)

QuantumBadger:
All working now and merged in. Thanks again!

PR #975, opened 2022-04-05
Add accessibility "concise mode" to make navigation more efficient for experienced Redditors

codeofdusk:
This PR implements the "concise" mode. Closes #971.

PR #977, opened 2022-04-07
Move accessibility "embellishments" before title in concise mode, adding read state as a post embellishment

codeofdusk:
This PR makes concise mode a bit more efficient by moving some "embellishments" (spoiler, sticky, NSFW, and flair) from after the title (in the accessibility subtitle) to before it (in the title) to ease rapid navigation. The reading order in standard (i.e. not concise) mode is unchanged.

I found a pair of eyes and answered my own question in https://github.com/QuantumBadger/RedReader/pull/975#issuecomment-1090762658. Apparently, submissions change colour in the list when read. This PR also adds the read state as an accessibility "embellishment", so it'll be read either before (concise) or after (standard) the title.

Note: this (and some of the other tricks employed in `org.quantumbadger.redreader.common.ScreenreaderPronunciation.getPronunciation`) are terrible for Braille, but Android's Braille support is a mess so my optimizing that in RedReader is low-pri. Ideally though, a "optimize for Braille" setting should be added that disables these hacks once Android gets decent Braille support or someone shouts.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/ScreenreaderPronunciation.java
+++ b/src/main/java/org/quantumbadger/redreader/common/ScreenreaderPronunciation.java
@@ -63,7 +63,7 @@ public class ScreenreaderPronunciation {
 		}
 
 		// Replace "read" with the English homophone "red" to work around bad speech synth handling
-		if(res == R.string.accessibility_read_withperiod) {
+		if(res == R.string.accessibility_post_already_read_withperiod) {
 			return "Red.";
 
 		} else {
--- a/src/main/java/org/quantumbadger/redreader/common/ScreenreaderPronunciation.java
+++ b/src/main/java/org/quantumbadger/redreader/common/ScreenreaderPronunciation.java
@@ -19,6 +19,7 @@ package org.quantumbadger.redreader.common;
 
 import android.content.Context;
 import androidx.annotation.NonNull;
+import androidx.annotation.StringRes;
 import org.quantumbadger.redreader.R;
 
 import java.util.Locale;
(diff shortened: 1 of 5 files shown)


codeofdusk:
CC @QuantumBadger.

codeofdusk:
@QuantumBadger I think it makes sense to have the "read" state occur **before** the post title in the accessibility description, so that users can easily swipe past already seen content.

To enable this, should I add a `BuildAccessibilityTitle` (that, for now, just concatinates the "read" string if needed and the post title), setting its result as the content description? Or do you suggest a different course of action?

QuantumBadger:
> should I add a BuildAccessibilityTitle

Thanks @codeofdusk, that sounds like a good idea.

Using the homophone "red" in strings.xml worries me a bit, because the comment above it won't show up for translators in Weblate, so I think everyone will just translate it as the colour. It is possible to manually add comments in the Weblate interface once the string has been committed, although it's the kind of thing that's easily ignored.

I've had a think about this, and it might be best to add a "pronunciation override" method like the following to the `ScreenreaderPronunciation` class. This method could be used for the relevant strings instead of calling `getString()` directly.

```java
	public static String getAccessibilityString(
			@NonNull final Context context,
			@StringRes final int res) {

		// Only override for English
		if(!Locale.getDefault().getDisplayLanguage().equals("en")) {
			return context.getString(res);
		}

		if(res == R.string.accessibility_subtitle_read_withperiod) {
			return "Red.";

		} else {
			return context.getString(res);
		}
	}
```

codeofdusk:
@QuantumBadger I think I've addressed your review actions, and added one more change now that we have a `buildAccessibilityTitle` to streamline concise mode further.

codeofdusk:
OK @QuantumBadger, I think this PR is now ready for review. I've also updated the description with some additional changes.

QuantumBadger:
Thanks @codeofdusk! I'll aim to get this reviewed and merged within the next couple of days.

QuantumBadger:
Nice work, thank you! I've merged this in and updated the changelog.

codeofdusk:
@QuantumBadger All working well, except:

* Perhaps add an entry to the changelog saying that read posts are now indicated to accessibility?
* The "read" -> "red" override is now not working for me. Suggestions on how to debug this?

PR #980, opened 2022-04-17
Fix-up of #977: fix  pronunciation of "read" state in English

codeofdusk:
This PR makes `getAccessibilityString` function properly by fixing the locale check, based on the usage described [in the Android documentation](https://developer.android.com/reference/java/util/Locale#getLanguage()).

Please add a changelog entry stating that read posts are now indicated in accessibility.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/ScreenreaderPronunciation.java
+++ b/src/main/java/org/quantumbadger/redreader/common/ScreenreaderPronunciation.java
@@ -58,7 +58,7 @@ public class ScreenreaderPronunciation {
 			@StringRes final int res) {
 
 		// Only override for English for now
-		if(!Locale.getDefault().getDisplayLanguage().equals("en")) {
+		if(!Locale.getDefault().getLanguage().equals(new Locale("en").getLanguage())) {
 			return context.getString(res);
 		}
 


codeofdusk:
CC @QuantumBadger. We're getting there! :)

codeofdusk:
CC @QuantumBadger. The diff is very small so review should be easy. We're getting there! :)

codeofdusk:
@QuantumBadger this PR is now ready for review. The diff is very small so review should be straightforward. We're getting there! :)

QuantumBadger:
Thanks @codeofdusk, I've merged this in, and also added a minor performance improvement, and a full stop + line break after the accessibility subtitle, so that the title and subtitle don't get merged into one sentence. Changelog also updated as requested.

Regarding the concise mode, I'm wondering if it would be less confusing to replace "u/user" in the subtitle with the original "by user", which takes about the same amount of time to say. The "u" sounds a lot like "you", which is confusing in the context of authorship :grin:

codeofdusk:
> Regarding the concise mode, I'm wondering if it would be less confusing to replace "u/user" in the subtitle with the original "by user"

I'd say not, because:

1. "u/" is more Reddit-like, and people who've been around know what that is.
2. A popular accessibility-minded client on iOS does it this way, so there's precedent.

Issue #962, opened 2022-03-14
Talkback has issues with long posts


davidbtc2009:
It seems as though talkback has issues with long reddit posts. I am unsure if it is due to the multiple links, or the post being long enough to go off the phone's display, or what. But it stops reading. I think it is actually the post going off screen and i will attach a link to an example post and screenshot.

The link https://www.reddit.com/r/degoogle/comments/tdumva/my_opinion_on_e_os_on_fairphone_4/
![Screenshot_20220314-133407_RedReader](https://user-images.githubusercontent.com/67274571/158229140-aef11d7d-830e-4242-97ba-a2bdc3fb5501.png)

From my screenshot you can see that the last few words are "this alternative launcher" and my talkback did not read anything after that point without intervention. When i do swipe it decides i have finished reading the post and begins going from link to link.

QuantumBadger:
Thanks for the report! It sounds like it might only be reading out paragraphs which are visible on screen.

If you go into Settings, Accessibility, and enable "Separate body text lines", does this fix the issue for you? It might be a good idea for us to enable this setting by default.

davidbtc2009:
Yes, that fixed it. However it also makes the links focusable through the links action rendering yhe link text and url boxes below the paragraphs as redundant.

codeofdusk:
Those can be disabled in settings -> appearance -> untick "show link buttons".

davidbtc2009:
Thanks for the help

Sent from ProtonMail mobile

-------- Original Message --------

PR #981, opened 2022-04-17
Separate body text lines by default

codeofdusk:
This PR changes the default value of "separate body text lines" to `true`.

You might also want to consider disabling "show link buttons" by default to address https://github.com/QuantumBadger/RedReader/issues/962#issuecomment-1069844482, but I haven't included that in this PR.


codeofdusk:
CC @QuantumBadger another small PR.

QuantumBadger:

> You might also want to consider disabling "show link buttons" by default to address https://github.com/QuantumBadger/RedReader/issues/962#issuecomment-1069844482, but I haven't included that in this PR.

Yeah, however we need to make sure this doesn't affect users who aren't using screenreaders. It might be best just to add that preference to the accessibility category, so it appears in both screens.

Issue #964, opened 2022-03-20
Add TalkBack accessibility actions


codeofdusk:
RedReader should have [custom accessibility actions](https://developer.android.com/reference/android/view/accessibility/AccessibilityNodeInfo.AccessibilityAction) to allow TalkBack users to quickly up/downvote posts, collapse threads, etc. similar to the assignable swipe gesture actions for sighted users. Ideally, the order of the actions would be customizable and users can choose which actions they'd like to have enabled, again similar to the already existing support for swipe gestures. The lack of an efficient, accessible Reddit client is one of the major things keeping me on iOS, so solving this would really mean a lot to me!

QuantumBadger:
Thank you for the suggestion, I'll look into this. Am I right in thinking that the swipe accessibility actions (ACTION_SWIPE_DOWN etc) are only supported on Android 13 onwards?

codeofdusk:
I'm not too sure and haven't heard of those before, but I imagine they'll be really useful when 13 launches!

QuantumBadger:
I've had a look into this. It seems to be possible add accessibility actions using `ViewCompat.addAccessibilityAction()`. The actions are accessible with TalkBack by:

* First, swiping up and right in an "upside-down-L" motion
* Selecting "Actions" from the menu that appears
* Selecting the specific action you want to perform from the next menu

Unfortunately this seems like a lot more work that the current double-tap-and-press to open the context menu. As far as I can see there isn't a way to add custom accessibility gestures for items.

@codeofdusk If there are any other apps which implement what you're looking for, could you send me a link please?

codeofdusk:
Yes, you found the actions menu I'm talking about.

You can assign a custom gesture to open the "actions" menu without having to find it in the LCM.

QuantumBadger:
@codeofdusk Thanks. I'm happy to add this, but just wanted to check first why the normal long click menu is inadequate. I'm guessing because (1) you can't change the order of the items, and (2) removing rarely-used items from it means they are totally unavailable.

codeofdusk:
@QuantumBadger If nothing else, I think we should have accessibility actions for moving to the previous/next parent comment. Those commands currently don't work from an accessibility perspective as, to activate them, it's necessary to move from the currently focused comment so focus doesn't know where to land (unless I'm missing something).

codeofdusk:
[...]

Issue #953, opened 2022-02-08
Feature request: Keyboard commands for common actions


devinprater:
Hello. I am a blind user of your app, so use a screen reader, like TalkBack or ChromeVox. While using the app on the ChromeBook, I'd love to be able to use keyboard commands to interact with the comment I'm on, like upvoting, downvoting, or replying. These keys seem to be open for use:

Control + a
: Upvote comment
Control + z
: Downvote comment
Control + r
: reply to comment

On Android, you could also use the accessibility [Actions menu](https://stackoverflow.com/questions/58092735/how-to-add-custom-actions-in-accessibility-local-context-menu) for Reply, upvote, downvote, all that, although the double tap and hold works well too.

codeofdusk:
See #964 as well for accessibility actions request.

codeofdusk:
@QuantumBadger might you be able to add this? A few thoughts:

* As not all devices with keyboards have <kbd>Ctrl</kbd> keys, I'm not sure that including <kbd>Ctrl</kbd> as a modifier key would be best.
* I'd consider this issue blocked by #986 (as in, it should probably be implemented first).

codeofdusk:
Also CC @Cguy7777 as I think he has a phone with physical keyboard.

PR #984, opened 2022-04-21
Move subreddit and domain before author in accessibility subtitle

codeofdusk:
For speech/Braille users, the most salient items should be placed first in the accessibility subtitle to ease interruptability and make navigation more efficient. To that end, this PR moves the subreddit and domain before the author in the accessibility subtitle.

While salience of these items is a bit subjective, I've discussed this with some other blind users who agree that subreddit and domain should be presented before author.

This probably doesn't deserve its own item in `changelog.txt` (can just be considered an "accessibility text improvement").

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditPreparedPost.java
@@ -1276,32 +1276,6 @@ public final class RedditPreparedPost implements RedditChangeDataManager.Listene
 					.append(separator);
 		}
 
-		if(mPostSubtitleItems.contains(PrefsUtility.AppearancePostSubtitleItem.AUTHOR)) {
-			@StringRes final int authorString;
-
-			if("moderator".equals(src.getDistinguished())) {
-				authorString = conciseMode
-					? R.string.accessibility_subtitle_author_moderator_withperiod_concise_post
-					: R.string.accessibility_subtitle_author_moderator_withperiod;
-			} else if("admin".equals(src.getDistinguished())) {
-				authorString = conciseMode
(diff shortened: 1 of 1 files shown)


codeofdusk:
@QuantumBadger This PR is ready for review. As it is simply moving code around, review should be straightforward.

Thanks again for merging all of my PRs, I'm extremely pleased with how RedReader's accessibility support is developing! 1.19 is looking to be a fantastic release for TalkBack users, both less and more experienced!

codeofdusk:
Thanks @QuantumBadger!

Just out of curiosity, any estimate on when 1.19 will be on the Play Store?

QuantumBadger:
Merged in. Thank you for all the contributions in this release @codeofdusk, they're much appreciated :)

I'm aiming to get v1.19 up soon, ideally within the next week, due an issue that cropped up with logging in via reddit. Still a handful of Google Play conformance stuff that needs to be done first, but nearly there.

codeofdusk:
@QuantumBadger It looks like the latest alpha on [the website](http://redreader.org/alpha) is 312 (the commit before this merge). Was 313 not published yet, or should I look for it elsewhere?

I'm really pleased with where things are, but a few more things would make the client perfect. In roughly descending priority order:

* https://github.com/QuantumBadger/RedReader/issues/964#issuecomment-1107633894 (partial accessibility actions support for moving among parent comments).
* #986 (confirmation of common actions via announcement events).
* #953 (keyboard shortcuts).
* #987 (user customization of accessibility description order).
* #964 (full implementation of TalkBack actions).
* #988 (Braille support).

Might you have some time to work on these, either for this or next release? (or some for this and some for next). Thanks again for your reviews and merges!

QuantumBadger:
@codeofdusk Alpha 313 is now built. Thanks for raising those other tickets, I'll aim to take a look at some of these over the next few weeks!

PR #1003, opened 2022-05-23
Add option to show comment count in post subtitle

### LINKED ISSUE #989: Optionally show number of comments in post subtitle
Per [this thread](https://www.reddit.com/r/RedReader/comments/tblskr/request_feature_post_subtitle_comment_count/#content), it'd be useful to show the number of comments on a post in the subtitle. This should be added before score in the accessibility description.

Cguy7777:
Several notes:
- I put this right after the score and upvote ratio, which struck me as a good spot.
- I used "cmts" for the visible text, since using the entire word "comments" adds a lot of length to the post subtitle, but maybe that's not a good abbreviation?
- I don't think the accessibility version could really be any more compact than `"%d comment(s)."`, so I always used that string.

Closes #989.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
+++ b/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
@@ -545,7 +545,18 @@ public final class PrefsUtility {
 	}
 
 	public enum AppearancePostSubtitleItem {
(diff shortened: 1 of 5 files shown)


QuantumBadger:
This looks good, thanks! Merged in and changelog updated.

I also can't think of a better abbreviation than "cmts" :) It should be fine I think.

PR #1004, opened 2022-05-23
Improve time and subtitle/header formatting with non-breaking spaces

Cguy7777:
Noticed the time periods, and a few subtitle/header entries, could use non-breaking spaces to occasionally improve readability.


QuantumBadger:
Looks good, thank you!

PR #1011, opened 2022-06-01
Add ability to hide subreddit header

### LINKED ISSUE #990: Hide "post list header"?
We have the option to disable the toolbar, but has there been consideration to hide the subreddit header and subcount as well?

The title is just duplicating the information already visible on the display near the back arrow at the top `<-/r/subreddit`. 

On my Pixel 4a with inline photo previews disabled, and the default preview icon size of 1.0 and default font size, this post header takes almost the same amount of space as a post, meaning less content on display at a time.

Cguy7777:
I thought I'd knock this one out, for the benefit of those with smaller screens (or who like to maximize the space for posts).

I'm not sure if it's 100% necessary to refresh when this preference is changed, but the very similar preference to hide the post header toolbar is also set to require a refresh. It *seems* like setting this might avoid some jank?

Closes #990.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
+++ b/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java
@@ -537,7 +537,7 @@ public final class PrefsUtility {
 	public static boolean pref_appearance_post_hide_subreddit_header() {
 		return getBoolean(
 				R.string.pref_appearance_post_hide_subreddit_header_key,
-				true);
+				false);
 	}
 
 	public static boolean pref_appearance_hide_headertoolbar_postlist() {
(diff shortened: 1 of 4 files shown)


QuantumBadger:
Thank you @Cguy7777! Looks good, although there's a bug where `pref_appearance_post_hide_subreddit_header()` defaults to `true`, but the XML preference defaults to `false`, which means that if someone hasn't visited the appearance preferences, this will default to hiding the header.

Cguy7777:
@QuantumBadger My bad, just accidentally set it wrong there. Should be all good to go now. :-)

QuantumBadger:
Perfect, thank you! Merged in and changelog updated.

PR #1029, opened 2022-09-06
Show user avatar in User Profile dialog

mgurga:
This commit adds the icon_img key to the RedditUser class as well as a helper function to strip problematic parts of the URL. I also created an avatar layout which is added to the User Profile dialog. The code to download the image and put it into the imageview is very hacky and I'm sure there is a better way to do it. If the URL returns some kind of error the avatar layout is never added.

Here's a screenshot of the final product:
image

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/fragments/UserProfileDialog.java
+++ b/src/main/java/org/quantumbadger/redreader/fragments/UserProfileDialog.java
@@ -32,7 +32,6 @@ import android.widget.TextView;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
-
 import org.quantumbadger.redreader.R;
 import org.quantumbadger.redreader.account.RedditAccountManager;
 import org.quantumbadger.redreader.activities.BaseActivity;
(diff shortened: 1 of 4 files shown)


QuantumBadger:
This is a nice idea, thank you! A couple of comments below:

* It would be good to center the avatar horizontally, and also make sure the bottom doesn't get cut off
* Normally RedReader downloads stuff using the `CacheManager`. This ensures that a proxy is used where necessary (e.g. for Tor), and that the image is cached to disk. See the following example below: https://github.com/QuantumBadger/RedReader/blob/e5113e9dd0027785e22f7d145f01902a1b6d65e6/src/main/java/org/quantumbadger/redreader/views/RedditPostView.java#L646-L655

mgurga:
Reddit does not have a limit on the size of avatars when you upload them, and sometimes they are very large and not square shaped. I just took the safe option and rendered them similar to how Reddit does when you hover over a user on old Reddit, so in this case it is supposed to be cut off (it's a different API token 'icon_img' vs 'snoovatar_img' in a user's about.json). 
image

QuantumBadger:
Thank you, I've made some minor changes and merged this in. I've also updated the changelog. Thanks again for working on this!

PR #1035, opened 2022-10-30
Add support for subreddit emotes

bharatknv:
Subreddit emotes are those :xxxx: (Where x is a single digit number) things you see in the comment sections of some subreddits. They're pretty niche and not supported by any reddit app except the official one as far as I know. They're used frequently in r/xqcow so you can take a look there if you want to test this out.

This is the first android project I've worked on so I might've made some dumb design decisions in my code. If you want me to change something, I'd be happy to give it a try.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/reddit/CommentListingRequest.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/CommentListingRequest.java
@@ -18,9 +18,11 @@
 package org.quantumbadger.redreader.reddit;
 
 import android.content.Context;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.UiThread;
+
 import org.quantumbadger.redreader.account.RedditAccount;
 import org.quantumbadger.redreader.account.RedditAccountManager;
 import org.quantumbadger.redreader.activities.BaseActivity;
@@ -46,6 +48,7 @@ import org.quantumbadger.redreader.reddit.prepared.RedditParsedComment;
 import org.quantumbadger.redreader.reddit.prepared.RedditParsedPost;
 import org.quantumbadger.redreader.reddit.prepared.RedditPreparedPost;
 import org.quantumbadger.redreader.reddit.prepared.RedditRenderableComment;
+import org.quantumbadger.redreader.reddit.things.EmoteMetadata;
 import org.quantumbadger.redreader.reddit.things.RedditComment;
 import org.quantumbadger.redreader.reddit.things.RedditPost;
 import org.quantumbadger.redreader.reddit.things.RedditThing;
@@ -54,6 +57,9 @@ import org.quantumbadger.redreader.reddit.url.RedditURLParser;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.Locale;
+import java.util.Map;
 import java.util.UUID;
 
 public class CommentListingRequest {
@@ -296,6 +302,43 @@ public class CommentListingRequest {
 		} else if(thing.getKind() == RedditThing.Kind.COMMENT) {
 
 			final RedditComment comment = thing.asComment();
+
+			if (comment.media_metadata != null && comment.media_metadata.asObject() != null) {
+				try {
+					final JsonObject metadataObj = comment.media_metadata.asObject();
+
+					final Iterator<Map.Entry<String, JsonValue>> jsonIterator = metadataObj
+							.iterator();
+
+					while (jsonIterator.hasNext()) {
+						final String key = jsonIterator.next().getKey();
+
+						final EmoteMetadata emoteMetadata = metadataObj
+								.get(key).asObject()
+								.asObject(EmoteMetadata.class);
+
+						// id is always structured as emote|{subreddit_id}|{emote_id}
+						// for subreddit emotes
+						if (emoteMetadata.id.split("\\|")[0] == "emote") {
+							final String emoteId = emoteMetadata.id.split("\\|")[2];
+
+							final String emotePlaceholder = String.format(Locale.getDefault(),
+									":%s:", emoteId);
+
+							final String imgTag = String.format(Locale.getDefault(),
+									"<img src=\"%s\" title=\"%s\"></img>",
+									emoteMetadata.s.u,
+									emotePlaceholder);
+
+							comment.body_html = comment.body_html.replace(emotePlaceholder, imgTag);
+						}
+					}
+				} catch (final Exception e) {
+					// Including this try-catch to cover for edge cases where reddit might send
+					// different values under media_metadata
+				}
+			}
+
 			final String currentCanonicalUserName = RedditAccountManager.getInstance(mContext)
 					.getDefaultAccount().getCanonicalUsername();
 			final boolean showSubredditName = !(mCommentListingURL != null
--- a/src/main/java/org/quantumbadger/redreader/reddit/CommentListingRequest.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/CommentListingRequest.java
@@ -18,11 +18,10 @@
 package org.quantumbadger.redreader.reddit;
 
 import android.content.Context;
-
+import android.util.Log;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.UiThread;
-
 import org.quantumbadger.redreader.account.RedditAccount;
 import org.quantumbadger.redreader.account.RedditAccountManager;
 import org.quantumbadger.redreader.activities.BaseActivity;
@@ -63,6 +62,8 @@ import java.util.Map;
 import java.util.UUID;
 
 public class CommentListingRequest {
+
+	private static final String TAG = "CommentListingRequest";
 
 	private final Context mContext;
 	private final CommentListingFragment mFragment;
@@ -344,6 +345,11 @@ public class CommentListingRequest {
 				} catch (final Exception e) {
 					// Including this try-catch to cover for edge cases where reddit might send
 					// different values under media_metadata
+					Log.e(
+							TAG,
+							"Exception while processing media metadata for "
+									+ comment.getIdAndType(),
+							e);
 				}
 			}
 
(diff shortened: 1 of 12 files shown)


QuantumBadger:
Thank you for this, and sorry for the delay in responding. I'll get this reviewed and merged soon!

src/main/java/org/quantumbadger/redreader/reddit/prepared/html/HtmlRawElementImg.java
```diff
@@ -0,0 +1,169 @@
...
+			element.getPlainText(stringBuilder);
+		}
+	}
+
+	public final synchronized void writeTo(
+			@NonNull final SpannableStringBuilder ssb,
+			@NonNull final AppCompatActivity activity) {
+		final int emoteLocationStart = ssb.length();
+
+		final CountDownLatch latch = new CountDownLatch(1);
```
QuantumBadger, line 79: It looks like this is blocking the thread until the download has completed. This is something we need to avoid, because:

* Some HTML parsing is done on the main thread, and if this blocks, then it could cause UI stuttering or ANRs (crashes)
* A comment section with lots of emotes would download these one at a time, so if for example we have 100 emotes, we'd need to block for 100 round-trip network requests before we show the comments, which could take several minutes on poor connections

Ideally we need to perform the downloads asynchronously in the background, in such a way that we can display the comments without waiting for the downloads to complete. If a download completes after the comment containing the emote is already displayed, then the comment should be updated.

This will require a bit of refactoring -- I'd suggest adding an "invalidate" callback as a parameter to `generate()`.

```java
public void generate(
			@NonNull final AppCompatActivity activity,
			@NonNull final ArrayList<BodyElement> destination,
			@NonNull final Runnable invalidateCallback)
```

Once the image was downloaded, `HtmlRawElementImg` would:

* Store the image in an `AtomicReference` member variable
* Invoke the invalidate callback

The code responsible for rendering the comment would then know to call `generate()` again. This time, the `HtmlRawElementImg` would already have the image in memory.

There would also need to be some changes to `HtmlReader.parse()` -- currently this:

* Parses the HTML
* Produces a reduced `HtmlRawElementBlock` by calling `reduce()` on the tree
* Produces a `BodyElementVerticalSequence` by calling `generate()` on the reduced block

That last step needs to be moved outside `parse()`, so that it can be run again (when the invalidate callback is invoked) without the need to do the parsing and reduction again (which would mean a new instance of `HtmlRawElementImg` would be created, and it wouldn't have the image cached in memory).

Hopefully that all makes sense! Happy to answer any questions about this.

src/main/java/org/quantumbadger/redreader/reddit/CommentListingRequest.java
```diff
@@ -296,6 +302,43 @@ private void buildCommentTree(
...
+
+							final String imgTag = String.format(Locale.getDefault(),
+									"<img src=\"%s\" title=\"%s\"></img>",
+									emoteMetadata.s.u,
+									emotePlaceholder);
+
+							comment.body_html = comment.body_html.replace(emotePlaceholder, imgTag);
+						}
+					}
+				} catch (final Exception e) {
```
QuantumBadger, line 336: We shouldn't ignore exceptions -- at the very least this should be logged, although typically we'd want to show a parse error next to the relevant comment so that the user can report this.

src/main/java/org/quantumbadger/redreader/reddit/prepared/html/HtmlRawElement.java
```diff
@@ -269,7 +269,9 @@ public static HtmlRawElement readFrom(@NonNull final HtmlReaderPeekable reader)
 					break;
 
 				case "img":
-					result = new HtmlRawElementPlainText("Image");
```
QuantumBadger, line 272: I think this might break GIFs in comments? Currently these are shown as a link containing this "Image" plain text element.

src/main/java/org/quantumbadger/redreader/reddit/CommentListingRequest.java
```diff
@@ -296,6 +302,43 @@ private void buildCommentTree(
...
+
+							final String emotePlaceholder = String.format(Locale.getDefault(),
+									":%s:", emoteId);
+
+							final String imgTag = String.format(Locale.getDefault(),
+									"<img src=\"%s\" title=\"%s\"></img>",
+									emoteMetadata.s.u,
+									emotePlaceholder);
+
+							comment.body_html = comment.body_html.replace(emotePlaceholder, imgTag);
```
QuantumBadger, line 333: This seems like a bit of an inefficient way to do this -- ideally this would be part of the parser itself. This is probably good enough though, since comments aren't typically that long and most comments don't contain any emotes.

QuantumBadger (changes requested):
Thanks, this mostly looks good! I've added some comments, particularly about how the images are downloaded.

src/main/java/org/quantumbadger/redreader/reddit/prepared/html/HtmlRawElementImg.java
bharatknv: Thanks for the feedback! I'll look into `HtmlReader.parse()` function when I have bit of time and see if I can make sense of your directions. 

I have just one question for now. Wouldn't it be simpler and more efficient to just update the textview in the callback?

src/main/java/org/quantumbadger/redreader/reddit/CommentListingRequest.java
bharatknv: Makes sense. I'll make it show a parse error to the user

src/main/java/org/quantumbadger/redreader/reddit/prepared/html/HtmlRawElement.java
bharatknv: I don't think it breaks gifs. I've been using my build for a few weeks now and haven't noticed anything. I'll double check just in case

src/main/java/org/quantumbadger/redreader/reddit/CommentListingRequest.java
bharatknv: So you're saying that this logic should be part of the JSON parser that deserializes the JSON into the reddit comment object? Or are you referring to the `HTMLReader` parser?

src/main/java/org/quantumbadger/redreader/reddit/prepared/html/HtmlRawElementImg.java
QuantumBadger: Thanks! Updating the `TextView` is probably fine too, it depends on exactly how it's implemented.

Feel free to implement it however you think is best and I'll take a look :)

src/main/java/org/quantumbadger/redreader/reddit/CommentListingRequest.java
QuantumBadger: I think it would fit best into the HTML parser, maybe as a "post processing" step (walk the element tree and split apart text elements which contain emotes). I think the current way is probably fine though.

QuantumBadger:
One other comment -- I think Reddit lets you use emotes in flair text too. That doesn't have to be done as part of this PR, but it would be great to support that too in future.

Thanks again for your work on this! Let me know if you have any more questions about the changes.

src/main/java/org/quantumbadger/redreader/reddit/prepared/html/HtmlRawElementImg.java
bharatknv: I made the emote download non-blocking. 

I created a callback in `BodyElementTextSpanned` that would update the `TextView` content based on the content of the `AtomicReference` variable. 

Not sure if this is the best way to do this or if updating the `TextView` could have side effects. Would appreciate your feedback

src/main/java/org/quantumbadger/redreader/reddit/CommentListingRequest.java
bharatknv: I'm logging this exception now. 

I don't think I've ever seen an exception be thrown here so I didn't bother with showing a parse error for this. I'll look into it if you think it's important

src/main/java/org/quantumbadger/redreader/reddit/prepared/html/HtmlRawElement.java
bharatknv: GIFs seem to work fine. The only side effect I've seen is that a preview of the GIF is now shown above the link, which is probably more of a feature than a bug

I'm now passing emotes with a custom `emote` tag just in case though. If we discover any bugs or users don't like the GIF preview, we can always revert this case condition in `HtmlRawElement.java`.

``` 
case "img":
	return new HtmlRawElementImg(children,
	        startToken.title == null || startToken.title.isEmpty()
			        ? "emote" : startToken.title,
	        startToken.src);
```
This will resolve it without affecting the subreddit emotes

bharatknv:
> One other comment -- I think Reddit lets you use emotes in flair text too. That doesn't have to be done as part of this PR, but it would be great to support that too in future.
> 
> Thanks again for your work on this! Let me know if you have any more questions about the changes.

Sure thing! I'll take a look at flair emotes and create a new PR if I figure it out. Can I create an issue for this? I could ask you questions in the issue comments if I run into any problems.

QuantumBadger:
Thank you for the changes!

I did some testing on this and got a lot of strange behaviour (particularly when tapping comments to collapse them) with comments being the wrong size, and not changing size when collapsed, I investigated this and it looks like the `SpannableStringBuilder` and `TextView` were being updated from non-UI threads.

I've added two commits with some fixes, which I think resolves all the problems I noticed. I need to have another read through the code at some point before merging this, and I'll do some more testing.

> I'll take a look at flair emotes and create a new PR if I figure it out. Can I create an issue for this? I could ask you questions in the issue comments if I run into any problems.

Yes that's fine, please do :)

> I noticed that Reddit never closes the img tags it sends in the body_html. Because of this, consecutive images are not being shown properly. To fix this, I changed the tag type for img tags to always be TAG_START_AND_END in HtmlReader.java

That's probably fine as long as Reddit never change the HTML they send us. I'll look into it a bit more when I do another check of this PR.

bharatknv:
I noticed the same strange behavior but couldn't figure out why it was happening. I need to read more about concurrency in Android. Thanks for fixing the issue!

QuantumBadger:
Thanks @bharatknv! I've merged this in and updated the changelog.

I made a few more minor changes, including turning off the emote precaching -- I think it would be bad for performance to parse the JSON for every comment section in the post list. Otherwise looks good, so thanks again!

PR #1043, opened 2023-01-22
'mark as read' fling option

JoshAusHessen:
**Changes:**
I would like to add the option 'mark as read' for the fling actions.
I also added it for the action menu.

**Background:**
I currently tip on images just to go back immediately bc. I want them to be hidden, when I refresh. That would be way easier with a fling option like that.

**Note:**
I hope I got everything I needed to change. Thanks for the great App. :)


JoshAusHessen:
:/ Ok I gonna investigate those errors later.

QuantumBadger:
Thank you! It just looks like a couple of minor syntax errors.

Issue #1023, opened 2022-07-08
Accessibility subtitle not read with ChromeVox


codeofdusk:
On playing a bit with RedReader on the Chromebook, it seemed like the accessibility subtitle wasn't being read (it says "pts" instead of "up votes" or "up", for instance). I don't have constant access to a Chromebook for further investigation/development but could borrow someone elses if needed.

@devinprater If I remember correctly, you use this app on ChromeOS. Does toggling "use concise language when possible" have any effect on the app's accessibility labels?

QuantumBadger:
Thanks for raising this @codeofdusk -- it's odd that it doesn't respect the `contentDescription`.

codeofdusk:
Given the move towards Aluminium OS, I'm not sure how much work ChromeVox's Android app support will be getting at this point. I'm going to close this unless someone strongly objects.

PR #1053, opened 2023-03-26
Add accessibility actions for jumping between level 1 comments

### LINKED ISSUE #964: Add TalkBack accessibility actions

codeofdusk:
This PR adds actions for jumping among level 1 comments, using the existing `previousButton`/`nextButton` logic. While it isn't quite my vision (collapsing the nearest l1), this will significantly ease navigation when it works properly.

While this does jump among l1s, focus is set extremely unreliably. @QuantumBadger might you be able to look into why?

Closes #964.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/fragments/CommentListingFragment.java
+++ b/src/main/java/org/quantumbadger/redreader/fragments/CommentListingFragment.java
@@ -754,11 +754,13 @@ public class CommentListingFragment extends RRFragment
 
 	private void setFocusDelayed(final int pos) {
 		AndroidCommon.UI_THREAD_HANDLER.postDelayed(() -> {
-			RecyclerView.ViewHolder view = mRecyclerView.findViewHolderForAdapterPosition(pos);
+			final RecyclerView.ViewHolder view
+					= mRecyclerView.findViewHolderForAdapterPosition(pos);
 			if (view != null) {
-				view.itemView.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED);
-				view.itemView.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_SELECTED);
-				view.itemView.performAccessibilityAction(
+				final View item = view.itemView;
+				item.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED);
+				item.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_SELECTED);
+				item.performAccessibilityAction(
 						AccessibilityNodeInfo.ACTION_ACCESSIBILITY_FOCUS,
 						null);
 			}
--- a/src/main/java/org/quantumbadger/redreader/fragments/CommentListingFragment.java
+++ b/src/main/java/org/quantumbadger/redreader/fragments/CommentListingFragment.java
@@ -26,6 +26,8 @@ import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.accessibility.AccessibilityEvent;
+import android.view.accessibility.AccessibilityNodeInfo;
 import android.view.animation.Animation;
 import android.view.animation.AnimationUtils;
 import android.view.animation.OvershootInterpolator;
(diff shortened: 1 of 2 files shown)


codeofdusk:
CC @QuantumBadger.

QuantumBadger:
Thank you @codeofdusk! I'll review this soon and will look into the focus issue.

QuantumBadger:
@codeofdusk After spending a while investigating this, I've put in a hack explicitly sets the accessibility focus to the jumped-to element, after a delay of 500ms to allow the `RecyclerView` to re-layout.

Please let me know if this works for you!

codeofdusk:
@QuantumBadger This does improve reliability, though not perfectly. Also, we're now getting #1023 type behaviour when invoking the actions – I find the other accessibility subtitle style more readable. That said, I imagine the new subtitle might be better for Braille users as it's more compact.

codeofdusk:
@QuantumBadger Any more thoughts on what to do here?

codeofdusk:
@QuantumBadger Glad to see that RedReader survives (at least for now)!

By #1023-style behaviour, I mean that TalkBack says things like "pts" instead of "up votes" or "up" and reads things in a more abbreviated style. That style works well for Braille (it's compact) but the normal accessibility subtitle works better for speech.

codeofdusk:
One possibility (suggested to me by @simon818) is that we mark all l1 comments as headings, so TalkBack's previous/next heading commands can move among them. Not sure how feasible this'd be, but I'd be okay with that solution.

QuantumBadger:
Thanks @codeofdusk, yes I'm glad the app's been given a stay of execution, in no small part thanks to the accessibility work! The "L1 comments as headings" idea seems like it might work, it looks like we can use the `setAccessibilityHeading()` method on the top-level comment views.

codeofdusk:
> The "L1 comments as headings" idea seems like it might work, it looks like we can use the setAccessibilityHeading() method on the top-level comment views.

It's been a while since I've looked at the app. Want to take this one, or should I? If the latter, is there an easy way to determine whether a comment is an l1?

QuantumBadger:
Happy to take a look at this myself in a moment! I should be able to use the indent level inside `RedditCommentView` for detecting L1 comments.

QuantumBadger:
Hmm, looks like it's more complicated than that unfortunately. Since the `RecyclerView` lazily generates the `View` instances, the view for the next "heading" doesn't exist yet, so there's nothing to jump to.

When I get a chance I'll take another look at the original PR and try to fix the remaining reliability issues.

codeofdusk:
@QuantumBadger Any updates on this?

QuantumBadger:
@codeofdusk No, I haven't had time to look into this further unfortunately. I'm still planning to take one more look to see if I can fix the original PR.

codeofdusk:
@QuantumBadger Any updates on this PR?

QuantumBadger:
@codeofdusk I think I'll merge this in as it's probably useful even if there are focus issues sometimes. As far as I can tell the focus issues are an Android issue and there's not much we can do to fix it. I put a workaround in place previously and it worked for me, and maybe the timeout can be tweaked to make it more reliable.

The issue with the accessibility subtitle also seems like an Android issue -- I'll do some more testing once this is merged in but I don't remember seeing this personally. In theory there's no reason why Android should pick a different item description depending on how you navigated to the item.

In the longer term:

* I'd like to implement a "collapse child comments" feature as this is often requested, and sidesteps the need for the extra skip actions
* The plan is to gradually migrate the app's UI to Jetpack Compose instead of Android Views, and in the upcoming version 1.24 the initial groundwork was done for this (with the album viewer). So I'd like to avoid investing much more time in the legacy Views-based UI. Compose handles accessibility differently and it's possible the issues we've seen here will no longer apply.

PR #1054, opened 2023-04-01
Add support for emotes in flairs

bharatknv:
Adds #1042

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/BetterSSB.java
+++ b/src/main/java/org/quantumbadger/redreader/common/BetterSSB.java
@@ -20,6 +20,7 @@ package org.quantumbadger.redreader.common;
 import android.graphics.Typeface;
 import android.text.SpannableStringBuilder;
 import android.text.Spanned;
+import android.text.TextUtils;
 import android.text.style.BackgroundColorSpan;
 import android.text.style.ForegroundColorSpan;
 import android.text.style.RelativeSizeSpan;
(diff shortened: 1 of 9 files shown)


bharatknv:
This is incomplete. I want to limit the emote height so it can't be taller than 2x the height of the text (or whatever multiple looks good).  

Currently I'm just halving the emote and it mostly works with my device and font size but that's obviously an imperfect solution.

@QuantumBadger , Any thoughts on what the best way to do this would be?

bharatknv:
The best solution I came up with is this:

```
final int TEXT_SIZE = 11; // Text size in sp
final float MAX_IMAGE_HEIGHT_MULTIPLE = 2.0F; // emote can only be 2x the size of the text


final float maxHeight = TypedValue.applyDimension(
  TypedValue.COMPLEX_UNIT_SP, 
  PrefsUtility.appearance_fontscale_comment_headers() * 
    TEXT_SIZE * 
    MAX_IMAGE_HEIGHT_MULTIPLE, 
  activity.getBaseContext().getResources().getDisplayMetrics());

if (image.getHeight() > maxHeight) {
      final float imageAspectRatio = (float) image.getHeight() / image.getWidth();
      
      final float newImageWidth = maxHeight / imageAspectRatio;
      
      image = Bitmap.createScaledBitmap(image,
		      Math.round(newImageWidth),
		      Math.round(maxHeight),
		      true);
}
```

I got the text size from reddit_comment.xml. I'd like to get it programmatically instead of hard-coding it but I don't see an obvious way to do that.

QuantumBadger:
Thank you for this @bharatknv and sorry for the very delayed response, there's a backlog of PRs that I'm just now clearing up!

> The best solution I came up with is this

That looks fine to me. I've adjusted `maxImageHeightMultiple` to 1.0, and also vertically centred the emotes. Will merge soon once the CI passes.

QuantumBadger:
Merged and changelog updated -- thanks again!

Issue #771, opened 2020-07-31
Broken mailto markdown detection


B0pol:
https://www.reddit.com/r/spacex/comments/gxb7j1/wearethespacexsoftwareteamaskusanything/

 At the bottom of the self-Text:
[…] send your resume to `[softwarejobs@spacex.com](mailto:softwarejobs@spacex.com)`

It should be displayed like that: [softwarejobs@spacex.com](mailto:softwarejobs@spacex.com) but it detects twice the emaid: once between `[ ]` (and it adds mailto), and another time between `( )`

<img alt="screenshot of the problem" src="https://i.ibb.co/p4LJWrz/Screenshot-20200731-192341-Red-Reader.png">

QuantumBadger:
Thanks for the report. It looks like that on the website too, though -- we don't actually parse markdown any more, we parse the already-formatted HTML :)

We could handle mailto links better in general though (and actually open the email app).

yawnoc:
> It looks like that on the website too [...] we parse the already-formatted HTML

The bad parsing only happens on Old Reddit:

````html
send your resume to [<a href="mailto:softwarejobs@spacex.com">softwarejobs@spacex.com</a>](mailto:<a href="mailto:softwarejobs@spacex.com">softwarejobs@spacex.com</a>)
````

![Old Reddit renders with incorrectly parsed mailto link](https://user-images.githubusercontent.com/48375680/105012732-6cedd280-5a79-11eb-9874-c68feb64c8a5.png)

The HTML output in New Reddit is fine:

````html
send your resume to <a href="mailto:softwarejobs@spacex.com" class="_3t5uN8xUmg0TOwRCOGQEcU" rel="noopener nofollow ugc" target="_blank">softwarejobs@spacex.com</a>
````

![New Reddit renders mailto link just fine](https://user-images.githubusercontent.com/48375680/105012882-9e669e00-5a79-11eb-9c5c-2eb179f6206a.png)

Could this be fixed by parsing New Reddit's HTML rather than Old Reddit's?

Cguy7777:
Currently working on this.

PR #1101, opened 2023-07-11
Several improvements with links/sharing/opening externally

### LINKED ISSUE #205: Add an option to open a post thread or comment in external browser
Would be great for things the client might not yet support (giving gold?). Thanks for RedReader!

### LINKED ISSUE #946: "share" / "save"  menu items
With the renaming of `share image...` to `share media...`  it'd like to suggest on additional small change for the share options

With the fact there are three share options, two of them (`share comments` and `share media`)  are very clear with what they do but the `share` on first look is not.   I'd like to suggest renaming this to `share link`.

I'd also like to suggest that the `Save Media` option be grouped beside the three `Share` menu items as they are quite similar in vein.

so the menu would look something like

    View in External Browser
    ....
    ....
    Save Media
    Share Link
    Share Comments
    Share Media
    Copy Link
    ....
    ....
    Properties

### LINKED ISSUE #947: [Feature Request] setting to share links as old.reddit.com
Please add a setting to RedReader → Settings → Behaviour → Sharing that makes shared links `https://old.reddit.com` instead of `https://reddit.com`.

### LINKED ISSUE #1099: Upper-case schemes in URLs do not open in external browser
In RedReader, URLs that appear in a comment (and likely a post as well) do not have the scheme filtered to lower-case (`https://` vs. `HTTPS://`). This is a problem as many apps do not receive intents where the URL scheme is in upper-case, resulting in RedReader opening the link in its own internal browser instead of the user's defined default browser, even if the app is set to open all links in an external browser. This may be undesirable because, for example:

* The internal browser will not be running extensions such as ad blockers,
* Any app designed to be an intermediary between apps and browsers (such as the excellent [URLCheck](https://github.com/TrianguloY/UrlChecker/)) will not catch the intent (though I've also opened [a feature request](https://github.com/TrianguloY/UrlChecker/issues/258) to allow URLCheck, at least, to catch these).

As an example, [this Reddit comment contains a link to `HTTPS://join-Lemmy.org`](https://www.reddit.com/r/ActLikeYouBelong/comments/14retl7/ractlikeyoubelong_needs_moderators_and_is/jqsi8bo/)). When opened in RedReader 1.21.1 with the "Behaviour->Use internal browser" option disabled, it fails to open the link in an external browser, and opens the link in its internal browser.

As schemes are defined as lower-case with a recommendation for programs to treat upper-case as if it was lower-case[1][2], I believe RedReader should automatically convert schemes in URLs to lower-case before opening the URL. (In this example, the link would be converted to `https://join-Lemmy.org` before attempting to open it.)

[1] https://www.rfc-editor.org/rfc/rfc1738#section-2.1
[2] https://www.rfc-editor.org/rfc/rfc3986#section-3.1

Cguy7777:
- URI hosts and schemes are normalized to lowercase. (#1099)
- Sharing, copying, and externally opening internal Reddit links works properly now. If someone made a Reddit link without the full URL ("r/test" rather than "https://reddit.com/r/test"), the "URL" sent out also did not contain the full URL. (previously unreported)
- Mailto links open properly in the app of choice for handling email. (partial fix for #771)
- Individual comments can be opened externally through a new option in the comment action menu. (#205)
- When sharing links, you can now choose which domain to share them from. Options are reddit.com (current behavior, default), redd.it for posts, Old Reddit, New Reddit, and np.reddit.com. Pre-made links using np.reddit.com are deliberately not changed, to respect their intention, but this can be changed if desired. (#947)
- When sharing a link to a post, if sharing titles is enabled, the title will be shared as the subject. In 3bea09f, behavior was changed so that the title would only be shared if the built-in share dialog was enabled, regardless of whether titles and descriptions were set to be shared, which seems to be an error. (previously unreported)
- Use the term "Share Link" for sharing post links, instead of just "Share", for clarity. (#946)

Closes #205, closes #946, closes #947, closes #1099.

### DIFF
--- a/src/main/assets/changelog-alpha.txt
+++ b/src/main/assets/changelog-alpha.txt
@@ -1,3 +1,9 @@
+/Alpha 336 (2023-07-09)
+Show self text for link posts (thanks to Cameron Merkel)
+Announce self-text presence with link posts in screen readers (thanks to Cameron Merkel)
+Add more specific error messages for blocked, quarantined, and premium-only subreddits (thanks to Cameron Merkel)
+Add option to disable user avatars (thanks to Cameron Merkel)
+
 /Alpha 335 (2023-07-02)
 Fixed an issue with subscriptions after re-authenticating
 Fixing crash when submitting a post
--- a/src/main/assets/changelog.txt
+++ b/src/main/assets/changelog.txt
@@ -1,4 +1,10 @@
-106/1.21.1
+108/1.22
+Show self text for link posts (thanks to Cameron Merkel)
+Announce self-text presence with link posts in screen readers (thanks to Cameron Merkel)
+Add more specific error messages for blocked, quarantined, and premium-only subreddits (thanks to Cameron Merkel)
+Add option to disable user avatars (thanks to Cameron Merkel)
+
+107/1.21.2
 Show Reddit user agreement prompt at startup
 Due to Reddit API changes, any active accounts must be logged in again
 More specific error message when a subreddit is private
--- a/src/main/java/org/quantumbadger/redreader/adapters/MainMenuListingManager.java
+++ b/src/main/java/org/quantumbadger/redreader/adapters/MainMenuListingManager.java
@@ -838,7 +838,10 @@ public class MainMenuListingManager {
 
 		switch(action) {
 			case SHARE: {
-				LinkHandler.shareText(activity, subredditCanonicalId.toString(), url);
+				LinkHandler.shareText(
+						activity,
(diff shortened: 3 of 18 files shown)


QuantumBadger:
Thank you for this @Cguy7777! I'll get it reviewed soon.

QuantumBadger:
I've had an initial look through and this is great work, thanks again for making all of these fixes. I'll do a final review pass soon and will get this rebased+merged.

QuantumBadger:
Thanks again! Merged and changelog updated.

PR #1107, opened 2023-07-15
Remember collapsed state of self-text

### LINKED ISSUE #937: RedReader does not remember the collapsed state for the text of self posts
I am using RedReader 1.18 from F-Droid. While RedReader remembers the collapsed state of the comments and replies under a post even across app restarts, it doesn't do so for the text of self posts.

Every time I open a self post, I have to collapse the text manually.

Cguy7777:
This should be good to go, but I suggest double checking this one particularly; there's work in part of the codebase I'm less familiar with.

Closes #937.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/fragments/CommentListingFragment.java
+++ b/src/main/java/org/quantumbadger/redreader/fragments/CommentListingFragment.java
@@ -84,7 +84,6 @@ public class CommentListingFragment extends RRFragment
 		CommentListingRequest.Listener {
 
 	private static final String SAVEDSTATE_FIRST_VISIBLE_POS = "firstVisiblePosition";
-	private static final String SAVEDSTATE_SELFTEXT_VISIBLE = "selftextVisible";
 
 	private final RedditAccount mUser;
 	private final ArrayList<RedditURLParser.RedditURL> mAllUrls;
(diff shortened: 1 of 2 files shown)


QuantumBadger:
Thank you for this @Cguy7777! Unfortunately there's a slight issue in that this changes the on-disk data format of the change data, which means the initial read fails when the app starts:

```
Initial read failed
java.io.EOFException
at java.io.DataInputStream.readFully(DataInputStream.java:205)
at java.io.DataInputStream.readUTF(DataInputStream.java:611)
at java.io.DataInputStream.readUTF(DataInputStream.java:566)
at org.quantumbadger.redreader.reddit.prepared.RedditChangeDataManager.readAllUsers(RedditChangeDataManager.java:161)
at org.quantumbadger.redreader.io.RedditChangeDataIO.runInitialReadInThisThread(RedditChangeDataIO.java:216)
at org.quantumbadger.redreader.RedReader$onCreate$2.run(RedReader.kt:73)
```

(this is because `Entry` tries to read an extra boolean field when it doesn't exist)

I think the cleanest solution to this is to change `RedditChangeDataIO.DB_VERSION` to `2`, and have `RedditChangeDataIO` pass the detected database version number into `RedditChangeDataManager.readAllUsers()`, which can in turn pass the version to the `Entry` constructor. `Entry` can then choose whether or not to read the field depending on the version. The `if(DB_VERSION != version)` check will also need to be changed or removed.

QuantumBadger:
Closing for now due to inactivity, but if this was fixed I'd be happy to merge.

PR #1110, opened 2023-07-19
Fix navbar color preference title

Cguy7777:
The title of the navbar color preference falsely states that it also controls the status bar color, despite the fact that it does not and never has. For reasons unknown I changed the string to say this in #710.

If it's desired, I could instead make the preference actually control both bars, or add a separate preference for the status bar, although I don't think there is very much demand for that.


QuantumBadger:
Merged, thank you!

PR #1123, opened 2023-08-30
Add support for opaque shared URLs, fixes #1122

### LINKED ISSUE #1122: New reddit thread sharing links don't work
These have the form of `https://reddit.com/r/<subreddit>/s/<shareCode>` and appear to be generated by the iOS app. If you send it a HEAD request, you'll be redirected to `https://www.reddit.com/r/<subreddit>/s/<shareCode>` and then the `Location` header in _that_ response (i.e. the `wwww.` prefixed req.) will be set to the actual real thread URI.

I don't know if only posts can be shared in this way, or if whole subreddits can be shared too.

Redreader does not handle these well at present and you just get a black screen. `RedditURLParser.parse` seems to just return back null

Logs show: 
```
wm_on_create_called: [28321549,org.quantumbadger.redreader.activities.LinkDispatchActivity,performCreate]
wm_add_to_stopping: [0,28321549,org.quantumbadger.redreader/.activities.LinkDispatchActivity,completeFinishing]
wm_restart_activity: [0,222850074,19049,org.quantumbadger.redreader/.activities.LinkDispatchActivity]
wm_set_resumed_activity: [0,org.quantumbadger.redreader/.activities.LinkDispatchActivity,minimalResumeActivityLocked - onActivityStateChanged]
wm_create_activity: [0,11381031,19049,org.quantumbadger.redreader/.activities.LinkDispatchActivity,android.intent.action.VIEW,NULL,https://reddit.com/...,268435456]
wm_pause_activity: [0,222850074,org.quantumbadger.redreader/.activities.LinkDispatchActivity,userLeaving=true,resumeTopActivity]
wm_finish_activity: [0,222850074,19049,org.quantumbadger.redreader/.activities.LinkDispatchActivity,app-request]
wm_on_create_called: [222850074,org.quantumbadger.redreader.activities.LinkDispatchActivity,performCreate]
wm_add_to_stopping: [0,222850074,org.quantumbadger.redreader/.activities.LinkDispatchActivity,completeFinishing]
wm_restart_activity: [0,11381031,19049,org.quantumbadger.redreader/.activities.LinkDispatchActivity]
wm_set_resumed_activity: [0,org.quantumbadger.redreader/.activities.LinkDispatchActivity,minimalResumeActivityLocked - onActivityStateChanged]
wm_create_activity: [0,220585280,19049,org.quantumbadger.redreader/.activities.LinkDispatchActivity,android.intent.action.VIEW,NULL,https://reddit.com/...,268435456]
wm_pause_activity: [0,11381031,org.quantumbadger.redreader/.activities.LinkDispatchActivity,userLeaving=true,resumeTopActivity]
wm_finish_activity: [0,11381031,19049,org.quantumbadger.redreader/.activities.LinkDispatchActivity,app-request]
wm_on_create_called: [11381031,org.quantumbadger.redreader.activities.LinkDispatchActivity,performCreate]
wm_add_to_stopping: [0,11381031,org.quantumbadger.redreader/.activities.LinkDispatchActivity,completeFinishing]
```

lol768:
This isn't very good quality code, but it _does_ seem to work for me on my device at least!

Particularly cringeworthy is the way in which I've ended up reaching around the whole `HTTPBackend` abstraction to get at the `okhttp` client directly. `RequestDetails` just isn't a flexible enough/low-level enough abstraction unfortunately for sending an unauthenticated HTTP HEAD request and pulling a response header out. I wanted to ensure it'd still use e.g. the SOCKS proxy settings though - which is why I'm not just using OKHttp directly.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/common/LinkHandler.java
+++ b/src/main/java/org/quantumbadger/redreader/common/LinkHandler.java
@@ -46,6 +46,7 @@ import org.quantumbadger.redreader.activities.WebViewActivity;
 import org.quantumbadger.redreader.cache.CacheRequest;
 import org.quantumbadger.redreader.fragments.ShareOrderDialog;
 import org.quantumbadger.redreader.fragments.UserProfileDialog;
+import org.quantumbadger.redreader.http.HTTPBackend;
 import org.quantumbadger.redreader.image.AlbumInfo;
 import org.quantumbadger.redreader.image.DeviantArtAPI;
 import org.quantumbadger.redreader.image.GetAlbumInfoListener;
@@ -230,9 +231,11 @@ public class LinkHandler {
 				case RedditURLParser.OPAQUE_SHARED_URL:
 					// kick off a thread to get the real url
 					new Thread(() -> {
-						final String realUrl = OpaqueSharedURL.resolveUsingNetwork(
+						final String toFetchUrl = OpaqueSharedURL.getUrlToFetch(
 								(OpaqueSharedURL) redditURL
-						);
+						).toString();
+						final String realUrl = HTTPBackend.getBackend()
+								.resolveRedirectUri(toFetchUrl);
 						if(realUrl != null) {
 							activity.runOnUiThread(() -> onLinkClicked(
 									activity,
--- a/src/main/java/org/quantumbadger/redreader/common/LinkHandler.java
+++ b/src/main/java/org/quantumbadger/redreader/common/LinkHandler.java
@@ -230,7 +230,9 @@ public class LinkHandler {
 				case RedditURLParser.OPAQUE_SHARED_URL:
 					// kick off a thread to get the real url
 					new Thread(() -> {
-						final String realUrl = OpaqueSharedURL.resolveUsingNetwork((OpaqueSharedURL) redditURL);
+						final String realUrl = OpaqueSharedURL.resolveUsingNetwork(
+								(OpaqueSharedURL) redditURL
+						);
 						if(realUrl != null) {
 							activity.runOnUiThread(() -> onLinkClicked(
 									activity,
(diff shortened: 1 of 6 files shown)


QuantumBadger:
Thank you for this! The CI failures look like some unnecessary imports.

It would be great to avoid breaking the `HTTPBackend` abstraction -- maybe the easiest solution is to move the bulk of the `resolveUsingNetwork()` function into `OKHTTPBackend` , by adding a new method to the `HTTPBackend` interface (`String resolveRedirectUri(String url)` perhaps?)

One other issue -- we need to make sure we're not doing synchronous network activity on the UI thread (which is where `onLinkClicked` is normally called from). It would be sufficient to make `resolveRedirectUri()` start a new thread, and give a callback with the result.

QuantumBadger:
Example link for testing this: https://reddit.com/r/BestofRedditorUpdates/s/stxMlEtK5H

lol768:
Thanks for the review @QuantumBadger

lol768:
> Thank you for this! The CI failures look like some unnecessary imports.
The CI should pass now :)
> 
> It would be great to avoid breaking the `HTTPBackend` abstraction -- maybe the easiest solution is to move the bulk of the `resolveUsingNetwork()` function into `OKHTTPBackend` , by adding a new method to the `HTTPBackend` interface (`String resolveRedirectUri(String url)` perhaps?)

Sure, have incorporated this suggestion and removed the ugly method to get the `mClient` directly.


> One other issue -- we need to make sure we're not doing synchronous network activity on the UI thread (which is where `onLinkClicked` is normally called from). It would be sufficient to make `resolveRedirectUri()` start a new thread, and give a callback with the result.

I _think_ this is already covered in L241 of `org.quantumbadger.redreader.common.LinkHandler`?

https://github.com/QuantumBadger/RedReader/blob/04226e5283ece72565f0f0520b45194f7e57eb4a/src/main/java/org/quantumbadger/redreader/common/LinkHandler.java#L241-L257

When I test it I don't get any `NetworkOnMainThreadException`s

Maybe I've missed something though!

QuantumBadger:
Perfect, thank you! I'll re-review this soon and get it merged in.

> I think this is already covered in L241 of org.quantumbadger.redreader.common.LinkHandler?

Ah you're right, I missed this on the original review. Thanks!

lol768:
Rebased

Issue #1022, opened 2022-06-27
Improvements to Find Subreddit


QuantumBadger:
- [X] Rename to "Find Location"
- [X] Add buttons for quickly jumping to subreddit, user, URL, or post search
- [ ] Fix issue with loading spinner sometimes not appearing
- [ ] Make Custom Location a separate menu item again
- [ ] Update migration: Disable "Custom Location" preference if "Find Location" is enabled
- [ ] Don't re-download subscriptions (re-use existing database)
- [ ] Hide subreddit description and subscriber count for subscriptions
- [ ] Include pinned and history subreddits in the search list
- [ ] Hide keyboard when the "Search" button is pressed

japanesephundroid:
@QuantumBadger whatever it is you did for this I think it sped up the whole app. 👍

Some issues I had with cache loading and the wheel spinning for a long time seem to be gone now. I will report back when I do some more testing but again this fix is great.

mattdbr:
Re the hiding keyboard improvement - is this not just a matter of adding a `searchView.clearFocus();` inside onQueryTextSubmit? 

Only problem would be that it doesn't hide keyboard if there's no input (guess depends what the desired behaviour here is)

dcdom2:
> Include pinned and history subreddits in the search list

Yes please. Prioritized and immediate (as you type)

PR #1126, opened 2023-09-03
Find Location: Hide keyboard when search is pressed and string queried (re #1022)

mattdbr:
Part fix for #1022 - particularly 'Hide keyboard when the "Search" button is pressed'

The only situation that this fix doesn't address is when there is no text in the SearchView (as onQueryTextSubmit() is not called). Could be fixed by extending the SearchView class: https://dev.to/foso/how-to-use-a-searchview-with-an-empty-query-text-submit-4afh, however feels overkill for an edge case (and frankly given the keyboard can be hidden with the back button in android, I cannot imagine many users intending to press the 'search' button on an empty string).


QuantumBadger:
Nice straightforward fix, thank you! I've merged this and updated the changelog.

Issue #1135, opened 2023-10-03
[Feature request] Add setting to not mark posts as read


qilto:
It would be nice if there was a setting to not mark any posts as read, even when you read posts.

PR #1145, opened 2023-11-25
Add setting for marking posts as read

qilto:
This PR adds a setting for marking posts as read, which is enabled by default. If the setting is disabled, then posts will not be marked as read, meaning that there will be no visual cue that the post is read and the post will not be hidden if "hide read posts" is enabled. Posts that were already marked as read will remain marked even after enabling this setting.

Closes https://github.com/QuantumBadger/RedReader/issues/1135.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditChangeDataManager.java
+++ b/src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditChangeDataManager.java
@@ -19,7 +19,9 @@ package org.quantumbadger.redreader.reddit.prepared;
 
 import android.content.Context;
 import android.util.Log;
+
 import androidx.annotation.NonNull;
+
 import org.quantumbadger.redreader.account.RedditAccount;
 import org.quantumbadger.redreader.account.RedditAccountManager;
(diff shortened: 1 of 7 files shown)


src/main/java/org/quantumbadger/redreader/reddit/prepared/RedditChangeDataManager.java
```diff
@@ -375,7 +376,7 @@ Entry markRead(final TimestampUTC timestamp) {
 					timestamp,
 					mIsUpvoted,
 					mIsDownvoted,
-					true,
+					PrefsUtility.pref_behaviour_mark_posts_as_read() && true,
```
QuantumBadger, line 379: This should respect the existing value of `mIsRead`:

```java
mIsRead || PrefsUtility.pref_behaviour_mark_posts_as_read()
```

(also, `x && true` is the same as `x` :))

qilto: Thanks, done!

QuantumBadger:
Thank you again for the PR! I've finally merged this in for the next release.

Issue #1057, opened 2023-04-13
[Feature Request] Add Pause & rewind functions to the video/gif player


Leonidius20:
Hello!
I would like to suggest adding a play/pause feature in the video/gif player, as well as a seekbar to rewind back and forward.
Thank you for your work!

decadent:
Already there, look in the options.

nascentt:
Settings>Images/Video>Enable Video Playback Controls

PR #1164, opened 2024-01-07
Playback speed control

folkemat:
To be able to change the speed of a video, this PR adds a video speed control setting.

It can be accessed via an icon/button in the playback control. I chose the clock (ic_time_dark) for this (because I couldn't find a more suitable icon). Maybe a dedicated icon would be better.

Tapping on the clock opens a material alert dialogue with a seekbar through which you can set the speed. I added four buttons (0.5x, 1x, 1.5x and 2x speed) as a fast setting underneath the seekbar. The slowest setting is 0.01 and the fastest is 3.0.

The current speed is displayed in brackets next to the time progress.

Demo picture:
[photo_2024-01-07_20-37-13](https://github.com/QuantumBadger/RedReader/assets/96705437/eb11b9b9-9ddc-425c-a3e7-4b084676fb94)

Finally, I set the default setting for displaying the Video Controls to true, because it seems to me to be an elementary thing and people are often confused that there are none (see e.g. #1057).

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -1902,4 +1902,9 @@
 	<string name="album_grid_pref_horizontal_padding">Horizontal padding</string>
 
 	<string name="album_pref_compact_title">Compact title</string>
+
+	<!-- 2025-03-12 -->
+	<string name="video_speed_setting">Change video speed</string>
+	<string name="video_speed_instruction">Choose playback speed</string>
+	<string name="video_speed_term">Speed</string>
 </resources>
--- a/src/main/assets/changelog-alpha.txt
+++ b/src/main/assets/changelog-alpha.txt
@@ -1,3 +1,6 @@
+/Alpha 358 (2025-03-12)
+Added video playback speed control (thanks to folkemat)
+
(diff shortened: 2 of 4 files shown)


src/main/res/values/strings.xml
```diff
@@ -1829,4 +1829,7 @@
 	<string name="pref_album_skip_to_first_key" translatable="false">pref_album_skip_to_first_key</string>
 	<string name="pref_album_skip_to_first_title">Automatically open first album image</string>
 
+	<!-- 2024-01-07 -->!
+	<string name="video_speed_setting" translatable="true">Change video speed</string>
```
QuantumBadger, line 1833: No need to specify `translatable="true"`

folkemat:
> Generally looks good, thank you! I'm a bit concerned about cluttering the video controls with too many options, but I might revisit the design for that soon.

I don't think the design looks too cluttered with this additional button now, but yes, it could be reworked at some point. 

-  I have removed the default setting change for the videocontrols 
- and reworked the strings.

QuantumBadger:
Thanks again! Merged in and changelog updated.

Issue #700, opened 2020-01-17
Follow Users


Monstanner:
Can you Guys implement an option to follow users? And if it doesn't cause too much work, can you add that users' posts appear in the timeline?

nascentt:
Adding this here as a workaround that's been found.
https://reddit.com/comments/matgxw

Silun:
Still can't follow users. Seems like a very essential feature.

PR #1166, opened 2024-01-09
Ability to follow and unfollow a user

### LINKED ISSUE #1156: [Feature Request] View & Subscribe to User Subreddit
Every user has a "u_<username>" subreddit, however, currently, there isn't an easy way to view/subscribe to this page.

folkemat:
Hi, this PR adds the ability to follow and unfollow a user.

- There is now a Follow / Unfollow button in the profile
- Since following is nothing more than subscribing to a subreddit called "u_username", followed users appear in the main menu under "Subscribed Subreddits" 
- Followed users get a "Followed" chip in their profile

Closes #1156 and #700

What do you think? I look forward to your feedback

### DIFF
--- a/src/main/assets/changelog-alpha.txt
+++ b/src/main/assets/changelog-alpha.txt
@@ -1,5 +1,6 @@
 /Alpha 359 (2025-04-20)
 Added support for emotes in comment flairs (thanks to bharatknv)
+Ability to follow/unfollow users (thanks to folkemat)
 Added "Mark as Read/Unread" fling action, and optional context menu item (thanks to JoshAusHessen and codeofdusk)
 
 /Alpha 358 (2025-03-12)
--- a/src/main/assets/changelog.txt
+++ b/src/main/assets/changelog.txt
@@ -1,5 +1,6 @@
 114/1.25
 Added video playback speed control (thanks to folkemat)
+Ability to follow/unfollow users (thanks to folkemat)
 Added support for emotes in comment flairs (thanks to bharatknv)
 Added "Mark as Read/Unread" fling action, and optional context menu item (thanks to JoshAusHessen and codeofdusk)
 Added preference to prevent posts being marked as read when clicked (thanks to Daniel Ho)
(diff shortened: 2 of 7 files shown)


folkemat:
I've noticed that if you follow a user using the "u_username" scheme, e.g. "u_spez", a subreddit "/user/spez" (which opens the profile) and one called "u_spez" (which actually shows the posts in "r/u_spez") are created: u_spez is removed when the main menu is refreshed, while /user/spez only appears after the refresh. 

No idea why that is, but I don't know the code much either, maybe you have an explanation @QuantumBadger? Or is it a Reddit thing? reddit.com/r/u_spez also automatically redirects to reddit.com/user/spez.

So..
- To subscribe you have to use the u_username scheme, because /user/username doesn't work. 
- To unsubscribe, you can use both schemas, but we have to test if we should use u_username or /user/username works, because who knows if u_spez was already upgraded to /user/spez or not.
- And to display the chip and the correct buttons we also have to check both for the same reason

I have now simply solved it by testing for both, but maybe it is better to somehow trigger an refresh after following?

folkemat:
I just noticed: The whole thing now has to be adapted to the block/unblock behaviour; no toast messages, but greyed out (un)follow loading buttons. I'll do that soon

QuantumBadger:
Thanks @folkemat, and sorry for the delay in responding. I've added some fixes and rebased this on the latest `master`.

> I've noticed that if you follow a user using the "u_username" scheme, e.g. "u_spez", a subreddit "/user/spez" (which opens the profile) and one called "u_spez" (which actually shows the posts in "r/u_spez") are created: u_spez is removed when the main menu is refreshed, while /user/spez only appears after the refresh.

The subreddit gets added to the local list in `onSubscriptionAttempt()` and `onSubscriptionAttemptSuccess()`, so to address this I assume we'd have to store the subreddit name as `/user/abc` in canonical form, and then rewrite it to `u_abc` at the point when we make the subscription API call in `subscriptionAction()`.

> I just noticed: The whole thing now has to be adapted to the block/unblock behaviour; no toast messages, but greyed out (un)follow loading buttons. I'll do that soon

That would be great, thanks!

QuantumBadger:
Closing this for now due to lack of activity

PR #1176, opened 2024-02-12
Preference to hide posts with few comments

mosteo:
This is pretty barebones but my Android coding skills are somewhat rusty.


QuantumBadger:
Thank you for this! The code looks good, I'll test before merging.

QuantumBadger:
Before merging this I think it would be good to improve the user experience a bit, in terms of disabling this feature. Currently by default, `Hide posts with few comments: -1` is shown in the settings if this is disabled. We should avoid using `-1` for this -- I think it would be better to split this into two preferences: one for enabling/disabling (a `CheckboxPreference`), and one for setting the numeric limit (which is [greyed out if the other preference is disabled](https://stackoverflow.com/questions/12415226/preferences-being-enabled-dependent-on-other-preference)).

QuantumBadger:
Closing due to lack of activity.

PR #1196, opened 2024-05-31
Update wrapper-validation-action and use setup-gradle action

equeim:


QuantumBadger:
Looks good, thank you!

Issue #827, opened 2020-12-21
Using device's gesture navigation, image viewer breaks after attempting back gesture


marcel263:
![redreader115glitch-edit](https://user-images.githubusercontent.com/76437842/102793511-ee324680-43a1-11eb-8edf-7bccb73db998.gif)

RedReader 1.15. updated Samsung device running Android 10 and OneUI 2.5 

Steps to reproduce

- Have Image viewer set to "Internal viewer (OpenGL)"
- Open any image in the app
- Attempt to make the back gesture - stop it by returning finger to the starting position
- Interact with the image, e.g. zooming or panning. 

Result: Image disappears

Expected:

 Image should not disappear

equeim:
@QuantumBadger This should be fixed now.

QuantumBadger:
Thanks @equeim! I'll close this ticket once I've updated the changelog.

equeim:
@QuantumBadger Also #1143 and #1157

QuantumBadger:
Thanks! I've closed the tickets.

marcel263:
Confirmed fixed, thank you for the work

PR #1197, opened 2024-05-31
Fix FingerTracker (#827)

equeim:
Handle ACTION_CANCEL event and add assertions in debug builds.


QuantumBadger:
Looks good, thank you!

PR #1199, opened 2024-06-02
Cleanup Android version checks

equeim:


QuantumBadger:
Thanks for this! I reverted the translation strings changes because this causes issues with Weblate.

PR #1203, opened 2024-07-04
Fix some deprecations

equeim:


QuantumBadger:
@equeim Great work, thanks for this! I've fixed a crash when logging in due to a changed intent extra type (String to Uri). Will get this merged shortly.

PR #1207, opened 2024-07-15
Add frame step buttons to internal video player

ecawthorne:
Hi! This PR is based on #716. It adds an option for stepping through videos approximately frame by frame. 

The setting is disabled by default and the new buttons are grayed out and disabled when playing as per this comment https://github.com/QuantumBadger/RedReader/pull/716#issuecomment-594867307 by @QuantumBadger 

> Thanks! This looks good to me, when I get a chance I'll test/merge it and maybe add new icons.
> 
> I think it would be good to keep the new buttons hidden by default and have a preference to enable them, but I'm happy to do this if it's not clear how.
> 
> It might also be worth greying out the buttons while the video isn't paused (which may require another set of icons in a different colour).

I apologize if I'm stepping on any toes, there's been no activity on #716 for over a year so I figured I'd pick it up. Any feedback or suggestions are very welcome if there's more that needs to be done.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/views/video/ExoPlayerWrapperView.java
+++ b/src/main/java/org/quantumbadger/redreader/views/video/ExoPlayerWrapperView.java
@@ -47,7 +47,6 @@ import androidx.media3.ui.PlayerView;
 import androidx.media3.ui.TimeBar;
 
 import com.google.android.material.dialog.MaterialAlertDialogBuilder;
-
 import org.quantumbadger.redreader.R;
 import org.quantumbadger.redreader.common.AndroidCommon;
 import org.quantumbadger.redreader.common.General;
@@ -146,7 +145,7 @@ public class ExoPlayerWrapperView extends FrameLayout {
 			addButton(createButton(
 					context,
 					mControlView,
-					R.drawable.icon_previous,
+					R.drawable.icon_restart,
 					R.string.video_restart,
 					view -> {
 						mVideoPlayer.seekTo(0);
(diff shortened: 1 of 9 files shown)


src/main/java/org/quantumbadger/redreader/views/video/ExoPlayerWrapperView.java
```diff
@@ -188,7 +186,66 @@ public ExoPlayerWrapperView(
...
+				if (PrefsUtility.pref_behaviour_video_frame_step()) {
+					final AtomicReference<ImageButton> stepBackButton = new AtomicReference<>();
+					final AtomicReference<ImageButton> stepForwardButton = new AtomicReference<>();
+					stepBackButton.set(createButton(
+							context,
+							mControlView,
+							R.drawable.icon_step_back,
+							R.string.video_step_back,
+							view -> {
+								mVideoPlayer.seekTo(mVideoPlayer.getCurrentPosition() - 33);
```
equeim, line 198: You can get frame rate of a video from getVideoFormat and calculate position from it.

Also you probably need to change SeekParameters (https://developer.android.com/reference/androidx/media3/exoplayer/SeekParameters) to NEXT_SYNC/PREVIOUS_SYNC depending on seek direction

ecawthorne: Thank you for the advice! I changed the calculation to use `getVideoFormat` as requested. I don't think we want to use `SeekParameters` here because that would change the seek behavior in such a way as to make every seek go to a key frame, which could be several seconds before/after where we actually want to seek to in this context. If that's not correct please let me know though.

equeim: I think you can just reset them back to default value immediately after calling seekTo

ecawthorne: The issue as I understand it is that the sync points used by SeekParameters aren't necessarily going to be (approximately) the next or previous frame when we're trying to step through a video, they could potentially be hundreds of frames in either direction. Can I ask what your reasoning is for wanting to use SeekParameters here?

equeim: I thought it would be desirable to seek the next/previous key frame here, but you are right, they could be seconds apart and that's not what we want here

src/main/java/org/quantumbadger/redreader/views/video/ExoPlayerWrapperView.java
```diff
@@ -188,7 +187,68 @@ public ExoPlayerWrapperView(
...
+					final AtomicReference<ImageButton> stepBackButton = new AtomicReference<>();
+					final AtomicReference<ImageButton> stepForwardButton = new AtomicReference<>();
+					stepBackButton.set(createButton(
+							context,
+							mControlView,
+							R.drawable.icon_step_back,
+							R.string.video_step_back,
+							view -> {
+								long frameRate = mVideoPlayer.getVideoFormat() != null ?  30 : (long) mVideoPlayer.getVideoFormat().frameRate;
+								mVideoPlayer.seekTo(mVideoPlayer.getCurrentPosition() - frameRate);
```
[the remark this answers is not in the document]
ecawthorne: You're absolutely right, I'm not sure what I was thinking there. I also unintentionally inverted the null check which hid a different issue: apparently `getVideoFormat()` is always null in this context. After looking into it a bit, it seems like it's due to the format being H.264/AVC which ExoPlayer doesn't supply a frame rate for.

The condition is fixed in 8e0efb3697e65c985850c9e2b89acba1fba65e01. If you have any suggestions on how to proceed for calculating the frame rate just let me know. I'm working on the assumption that there possibly be a format change at some point so I kept the format check there for now.

QuantumBadger:
Thanks for this, and sorry for the delay with reviewing! I've rebased this and made a few minor adjustments. With the playback speed control, there are now too many icons to fit on the screen horizontally, so I'll have a think about how to improve the layout. I'll merge this in the meantime -- thanks again!

PR #1210, opened 2024-07-20
Mark all inbox messages as read when leaving the inbox

folkemat:
I found it annoying to constantly have to press ‘Mark all as read’ in the inbox menu. So this PR adds the option to automatically mark all messages as read when you leave the inbox, toggled via a checkbox in the inbox menu.


QuantumBadger:
Thanks for this! In general it looks fine, although there are potentially a couple of issues (that also exist with the current "mark all as read" option):

* If the user receives more messages after loading the page (but before going back), those messages will be marked as read despite never being seen
* Messages which are lower down in the list which weren't seen will also be marked as read

I'm wondering if it might be better to send an `/api/read_message` request containing only the messages which were actually shown on the user's screen (i.e. when the user scrolls we update a Set, and then when going back we send the request). We could also update the existing menu item to only mark the viewed items as read.

Would be interested to hear your thoughts!

folkemat:
I agree with you, these are valid points. I have implemented the API/read_message function in https://github.com/QuantumBadger/RedReader/pull/1238 for individual messages that are tapped on, it's probably best to merge that PR first

QuantumBadger:
@folkemat Thank you, I'll take a look at that PR.

As a sidenote -- this PR marks all the messages as read regardless of the setting :)

PR #1223, opened 2024-08-22
Fix triggering of single/double tap action when performing system navigation gesture

equeim:
Screen recording of issue that this patch fixes:

https://github.com/user-attachments/assets/7b694585-bc48-4269-92ff-70f1bdf55f52

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/views/imageview/BasicGestureHandler.java
+++ b/src/main/java/org/quantumbadger/redreader/views/imageview/BasicGestureHandler.java
@@ -71,7 +71,7 @@ public class BasicGestureHandler
 	}
 
 	@Override
-	public void onFingerUp(final FingerTracker.Finger finger) {
+	public void onFingerUp(final FingerTracker.Finger finger, final boolean cancelled) {
 
 		mCurrentFingerCount--;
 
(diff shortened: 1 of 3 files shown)


equeim:
Also discovered another issue when swipe overlay (when viewing image from a gallery) can get stuck when doing a very short swipe. The reason is that mCurrentTouchState may not be changed to ONE_FINGER_DRAG due to condition here: https://github.com/QuantumBadger/RedReader/blob/f7c9b90e3ebd022e90faf49bfd0a424dd6b8bfd3/src/main/java/org/quantumbadger/redreader/views/imageview/ImageViewDisplayListManager.java#L436, but because of fallthrough the onHorizontalSwipe listener is called anyway causing the overlay to show. However onHorizontalSwipeEnd listener that hides the the overlay is called only if state is ONE_FINGER_DRAG: https://github.com/QuantumBadger/RedReader/blob/f7c9b90e3ebd022e90faf49bfd0a424dd6b8bfd3/src/main/java/org/quantumbadger/redreader/views/imageview/ImageViewDisplayListManager.java#L528C15-L528C35, which means it won't be called in this case.

Not sure how to fix this properly without making this code even messier

Issue #1221, opened 2024-08-20
In v1.24 tapping certain URLs yield the "malformed URL" error page even though the URLs are valid


kappa7194:
Tapping certain links in comments lead to the "malformed URL" error page being shown even though the URLs are valid.

It seems to happen mainly for `i.imgur.com` URLs?

You can use nearly any post in r/photoshopbattles to reproduce the problem.

e.g.

https://old.reddit.com/r/photoshopbattles/comments/1evhn4i

https://old.reddit.com/comments/1evhn4i/comment/liscede

https://i.imgur.com/YvLcAcz.png

The full error is:

```
Error report -- RedReader v1.24

Manufacturer: samsung
Model: SM-G781B
Product: r8qxeea
Android release: 13
Android SDK: 33

-------------------------------
Title: Malformed URL
Message: An attempt was made to access an invalid URL.
URL:
Exception: java.lang.IllegalArgumentException
Expected URL scheme 'http' or 'https' but no scheme was found for
  okhttp3.HttpUrl$Builder.parse$okhttp(SourceFile:1093)
  okhttp3.Request$Builder.url(SourceFile:65)
  org.quantumbadger.redreader.http.okhttp.OKHTTPBackend.prepareRequest(SourceFile:494)
  org.quantumbadger.redreader.cache.CacheDownload.<init>(SourceFile:56)
  okhttp3.internal.cache.CacheStrategy.add(SourceFile:4)
  org.quantumbadger.redreader.RedReader$onCreate$1.queueDownload(SourceFile:44)
  org.quantumbadger.redreader.RedReader$onCreate$1.handleRequest(SourceFile:66)
  org.quantumbadger.redreader.RedReader$onCreate$1.run(SourceFile:18)
```

kappa7194:
The settings for the viewera in "Images/Video" must be set to "internal" to reproduce the problem.

The settings for "Use internal browser" and "Use Android custom tabs" in "Behavior" do not matter.

folkemat:
It's strange because I'm very sure that the links worked yesterday in 1.24. They still work in 1.23. I'm not sure if it has something to do with the recent okhttps changes or if Imgur changed something, or both together

folkemat:
Not sure why, but changing the Imgur pattern in LinkHandler.kt from

`"https?://?(i\\.)?imgur\\.com/(\\w+).*")`

to

`"http(s)?://?(i\\.)?imgur\\.com/(\\w+).*")`

fixes the issue.

[...]

PR #1224, opened 2024-08-22
Fix Imgur issue (#1221)

equeim:
"mp4" value can be present but empty.

Also handle this for "link" just in case.


QuantumBadger:
Nice work, thank you for looking into this! I've updated the changelog.

Issue #646, opened 2019-01-17
[Feature request] blacklist post titles


251:
It would be nice to have a blacklist of strings for post titles, e.g. "cat", "dog", "mods are asleep", ... Nothing fancy (regex), just plain case-insensitive search.

Hope that's not a duplicate - I could only find requests for tag/url-based filters. Thanks for the nice project.

thousandsofthem:
Totally with you, so many use-cases

Rub-z:
I'm also interested in this potential feature. Are there any key devs reading this post that could point me in the right direction? I.e. what class(es) would need to be changed to accommodate?

Thanks for taking the time to read and hopefully reply :)

QuantumBadger:
@Rub-z There are two major parts to implementing this, which are:

1. Make some suitable UI to set the preference
2. Filter posts before they're displayed based on that preference

(2) is fairly easy, the relevant class is `PostListingFragment`. Care needs to be taken to make sure this doesn't break related features like blocking subreddits or restricting post count.

https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java

In terms of the UI, it would be good to have a list where you can add/edit/delete entries, similar to the following random example off Google Images :)

![movie-list-android-56b754978fce492ea4882557e03fa8bf4402ed2036ea80d056b60e6c0e0b339c](https://user-images.githubusercontent.com/2998158/94342007-06deca80-0006-11eb-9e3a-e6fc7ce6d37c.png)

Relevant files are:

https://github.com/QuantumBadger/RedReader/blob/master/src/main/res/xml/prefs_behaviour.xml

https://github.com/QuantumBadger/RedReader/blob/master/src/main/java/org/quantumbadger/redreader/common/PrefsUtility.java

If you have any questions let me know :)

Rub-z:
Hi QuantumBadger! 

Many thanks for the insights :)  

I'd worked out that the PostListingFragment was key but this additional information is brilliant!

In regards to (1) the UI, is Settings > Behaviour > Posts a good place? 

Another option, 'Post Filtering', could be added which would direct you to a list of strings as per your image. 

I won't be able to properly look at the codebase until next week but will definitely take you up on any offer of advice for sure when I do! 

Many thanks for replying :)

QuantumBadger:
No problem :) Thanks for looking into implementing this!

[...]

PR #1226, opened 2024-08-25
Filter posts

eikaramba:
this fixes https://github.com/QuantumBadger/RedReader/issues/646 and allows to specify a comma separated list of keywords which will be used to filter posts by title.

In the future of course regex or filter based on user flairs can be integrated.

### DIFF
--- a/src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
+++ b/src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
@@ -100,6 +100,7 @@ import org.quantumbadger.redreader.views.liststatus.ErrorView;
 import java.io.IOException;
 import java.text.NumberFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.UUID;
@@ -812,6 +813,10 @@ public class PostListingFragment extends RRFragment
 									&& mPostListingURL.asSubredditPostListURL().type
 									== SubredditPostListURL.Type.SUBREDDIT);
 
+							final String keywordFilter = PrefsUtility.pref_behaviour_posts_filter();
+
+							String[] keywordFilterArray = keywordFilter.split(",");
+
 							final ArrayList<RedditPostListItem> downloadedPosts
 									= new ArrayList<>(25);
 
@@ -837,6 +842,15 @@ public class PostListingFragment extends RRFragment
 								final boolean isPostBlocked = subredditFilteringEnabled
 										&& blockedSubreddits.contains(
 										new SubredditCanonicalId(post.getSubreddit().getDecoded()));
+
+								String postTitle = post.getTitle().toString().toLowerCase();
+								if (!keywordFilter.isEmpty()) {
+									boolean keywordMatched = Arrays.stream(keywordFilterArray)
+											.anyMatch(keyword -> postTitle.contains(keyword.toLowerCase()));
+									if (keywordMatched) {
+										continue; // Continue the outer loop
+									}
+								}
 
 								if(!isPostBlocked
 										&& (!post.getOver_18() || isNsfwAllowed)
--- a/src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
+++ b/src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
@@ -815,7 +815,7 @@ public class PostListingFragment extends RRFragment
 
 							final String keywordFilter = PrefsUtility.pref_behaviour_posts_filter();
 
-							String[] keywordFilterArray = keywordFilter.split(",");
+							final String[] keywordFilterArray = keywordFilter.split(",");
 
 							final ArrayList<RedditPostListItem> downloadedPosts
 									= new ArrayList<>(25);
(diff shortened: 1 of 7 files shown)


src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
```diff
@@ -838,6 +843,17 @@ public void onDataStreamComplete(
 										&& blockedSubreddits.contains(
 										new SubredditCanonicalId(post.getSubreddit().getDecoded()));
 
+								final String postTitle = post.getTitle()
+										.toString().toLowerCase(Locale.US);
```
QuantumBadger, line 847: This should be moved inside the if statement

src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
```diff
@@ -812,6 +813,10 @@ public void onDataStreamComplete(
 									&& mPostListingURL.asSubredditPostListURL().type
 									== SubredditPostListURL.Type.SUBREDDIT);
 
+							final String keywordFilter = PrefsUtility.pref_behaviour_posts_filter();
+
+							final String[] keywordFilterArray = keywordFilter.split(",");
```
QuantumBadger, line 818: The keywords should be trimmed of whitespace here. It would also be good to move the `toLowerCase` on line 851 here to avoid doing it for every post.

src/main/res/values/strings.xml
```diff
@@ -82,6 +82,13 @@
     <string name="pref_behaviour_nsfw_key" translatable="false">pref_behaviour_nsfw</string>
     <string name="pref_behaviour_nsfw_title">Show NSFW content</string>
 
+	<string name="pref_behaviour_posts_filter_key" translatable="false">pref_behaviour_posts_filter</string>
+	<string name="pref_behaviour_posts_filter_header">Posts</string>
```
QuantumBadger, line 86: Not used anywhere

QuantumBadger:
Thanks for this! I've left a few minor comments, although I'm not sure I'd want to merge this in its current form.

Currently it's just doing a `title.contains(keyword)` to check if it matches, but that's not really sufficient as the keyword could be part of some other word in the title. For example, adding the word "trump" to the filter will filter out the word "trumpet", "trumping", etc. In other words this falls foul of the [Scunthorpe problem](https://en.wikipedia.org/wiki/Scunthorpe_problem).

I guess we'd want to split titles by punctuation/whitespace and check for matches that way?

src/main/java/org/quantumbadger/redreader/fragments/PostListingFragment.java
```diff
@@ -838,6 +843,17 @@ public void onDataStreamComplete(
 										&& blockedSubreddits.contains(
 										new SubredditCanonicalId(post.getSubreddit().getDecoded()));
 
+								final String postTitle = post.getTitle()
```
QuantumBadger, line 846: PostListingFragment is already quite cluttered -- it would be nice to split out all this new code into a new class.

thousandsofthem:
> I guess we'd want to split titles by punctuation/whitespace and check for matches that way?

It could be a phrase, like long series/book/etc title

eikaramba:
thank you for having a look at the PR. yes you are absolutely right, i changed it so that the keywords are matched against complete words (using \\W+, which matches one or more non-word characters (punctuation, whitespace, etc.))

I also resolved the aforementioned reviews apart from the refactoring, as i am unsure how you would like this to be done. i guess it would be good to refactor the whole class then and also make all the other conditions separate methods/classes.

folkemat:
Very nice. I see you put the option in Settings -> Behaviour -> Posts. I would instead suggest creating a whole new settings item, Settings -> Filters. Here we could then simply add more filters later (by Post domain, keyword, and Comment keyword etc. pp). Makes it easier to find and access in my opinion.

eikaramba:
no problem i modified the PR tp introduce a new settings category

eikaramba:
i think this can be merged no?

QuantumBadger:
While I appreciate the contribution, I've decided to close this PR as I'm not happy with most aspects of the implementation:

* UI: I don't think that a text box with a comma separated list is a good user experience. I'd like to make a proper list editing UI for this in Compose.
* Storage: shared prefs aren't a great choice for this -- we've run into issues with blocked subreddits where the pref string gets truncated due to being too long. I'd like to store this in an actual database, and I'm working on some changes which will improve how data is persisted in RedReader in general.
* Filtering mechanism: I think this needs some more thought (and maybe more flexibility for the user).
* Inefficiency: this is a minor point, but the code in this PR is needlessly inefficient (repeatedly calling `Arrays.asList(titleWords)` for every keyword in the list even though `titleWords` doesn't change, repeatedly calling `keyword.trim()` for every post/keyword combination even though the list of keywords doesn't change, leading to `n^2` string allocation for no reason, etc). 

Again, I'm grateful for the PR, but I'd rather implement this properly first time rather than have to redo it in future and migrate the preference data. I know that lots of people want this feature, so I'm planning to work on it myself, making use of the upcoming storage changes, and the new Compose UI work.

eikaramba:
you raise some valid points so it is understandable. i hope you can reuse some of the ui elements at least so it was not for nothing :)

GiraffeCubed:
I've been using this and my Reddit experience has been so much better. Looking forward to seeing it in an official capacity but thanks for your work on this, being able to filter stuff out is a godsend.

AXiSource:
This would be useful in the next "REDACTED said this" , "REDACTED said that" few years.

eikaramba:
last time any activity from @QuantumBadger was when he replied to this. it's nearly 6 months now, hopefully everything is alright. i also wish this would be implemented, maybe i will need to use my own fork from now.

PR #1251, opened 2024-11-25
Add Subreddit to Multireddit

mrboisvert:
@QuantumBadger awesome app, I've been using it for ages. Thought I would take a shot at adding a multireddit management feature that I've wanted for a while.

This MR adds a subreddit action 'Add to Multireddit'. Users can add a subreddit to a multireddit, or create a multireddit while adding a subreddit to it at the same time. This also does some refactoring to support making PUT requests, which I used to implement the 'Add to Multireddit' action.

The endpoint for adding a subreddit to a multireddit returns 201 on success and 400 if an illegal multireddit name is used. This MR also adds handling for these status codes.


mrboisvert:
There is also a commit in here refactoring CacheRequest to use a builder. I can drop it if it doesn't seem necessary.

PR #1257, opened 2024-11-29
Create Multireddit from Subreddit Combination

mrboisvert:
When viewing a subreddit combination (i.e. /r/sub1+sub2), user can long tap on the header and select `Create Multireddit` to create a multireddit out of all of the subreddits they are viewing.


QuantumBadger:
@mrboisvert Thank you for all the PRs, and sorry for the slow review. There's a backlog at the moment but I'm hoping to look at them soon.

PR #1260, opened 2024-12-02
Always preview

edgan:
I added an `Always Preview Mode` to always display previews instead of thumbnails. I prefer the consistency this provides. The default is `false`.

I tested toggling it `off` and `on` repeatedly on my `Google Pixel 8 Pro` running `Android 15`.

I did the translations for languages that I could find the `prefs_category_images_video` string for. I  double checked the translations. Added dates to the translation additions to follow what seem be a convention.


edgan:
I am closing this, because I have created another PR that is a superset of this one.

PR #1262, opened 2024-12-06
Layout mode

edgan:
In my first PR I added an Always Preview Mode to always display previews instead of thumbnails. I prefer the consistency this provides. 

In this second PR I added a variation of it call Card. It moves everything but the title to after the image. Based on how at least I scroll it makes the actions more accessible given the larger images of Always Preview Mode.

I tested toggling between all three modes repeatedly. I also made sure the layout was as expected per mode. I did this testing on my Google Pixel 8 Pro running Android 15.

I did the translations for languages that I could find the `prefs_category_images_video` string within the file for that language. I double checked the translations. Added dates to the translation additions to follow what seem be a convention.

[Example screenshots](https://imgur.com/gallery/redreader-layout-mode-5AzrOHs)

Modes:
```
Thumbnails
Always Preview
Card
```


QuantumBadger:
@edgan Thank you for this, and sorry for the slow review time. This looks interesting and I'll get around to it soon.

jacklollz2:
Yes please! This is just what I'm looking for!

Issue #1263, opened 2024-12-07
Reddit has removed the /r/random/ feature


FliegendeWurst:
When I try to visit a random sub via search for "random", the app claims it is banned.

```
{"reason": "banned", "message": "Not Found", "error": 404}
```

mnmnk43434:
Hello, we tried to solve the issue.

This is what we did:

Modified the SubredditPostListURL class to properly handle the API response for random subreddits. Added a new method to process the response and update the URL accordingly.

You can review changes in this commit: https://github.com/mnmnk43434/QuantumBadger-RedReader-1263/commit/78688e7393fe8890a5db73e0e094296995f28162.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

FliegendeWurst:
I am impressed, that looks like it might just work.

FliegendeWurst:
Nevermind, it is badly merged and uses imaginary methods..

FliegendeWurst:
Hmm.. visiting /r/random in the browser doesn't work either.

QuantumBadger:
Unfortunately Reddit has removed /r/random: https://old.reddit.com/r/help/comments/1fojw02/cleaning_up_some_lowusage_features/

I'll leave this ticket open to also remove the feature from RedReader, and also I've blocked the AI spam account which posted above :laughing:

pullpush-io:
Hi! I'm scraping Reddit in real time, and I'm just doing a major rewrite. Would you like me to add an endpoint for that to pullpush?

PR #1271, opened 2025-02-02
Remove random / randomNSFW shortcuts

vafada:
fixes https://github.com/QuantumBadger/RedReader/issues/1263

removes the random and random NSFW shortcut since https://old.reddit.com/r/random is now a banned subreddit making this feature useless


QuantumBadger:
Thanks @vafada, sorry for the slow review!

Mostly looks good, but we should avoid changing the non-English `strings.xml` files, as that causes merge conflicts with Weblate.

vafada:
@QuantumBadger pr updated

android studio is messing with new line at the end of the file on some non-english xml

is that okay?

QuantumBadger:
Thanks @vafada! I fixed a crash which occurred if the user had these options enabled.

PR #1291, opened 2025-05-15
ci: improve workflow

ArthurKun21:
- updated the setup_gradle and gradle wrapper validation to v4
- separate the pull request and push event trigger
- separate the assemble and the code checks
- added always run all of the code check even if previous step fails
- remove sdk and ndk and depend on the runner image for the up to date version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,59 +1,74 @@
 name: CI
 
 on:
-    workflow_dispatch:
-    push:
-    pull_request:
-    schedule:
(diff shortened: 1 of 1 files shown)


QuantumBadger:
Thank you for this. I've taken a look and there are several things in here that would be regressions, for example not installing the version of the NDK pinned in build.gradle.kts, the other tasks continuing if Gradle validation fails, the artifact renaming, etc. So on balance I'd rather not merge this, but nonetheless I appreciate the contribution!

PR #1298, opened 2025-06-18
Settings to filter out posts with few comments

mosteo:
This is another attempt at #1176, with the feedback that was given there.

When the checkbox is enabled, the user is asked for the minimum number of comments. This info is shown as the summary of the checkbox, when it is enabled. This way there are no visible magic numbers or values with special significance.

A toast is shown when comments are filtered, just as a reminder that this filter is in effect.

Caveat: while I did #1176 entirely the old fashioned way, I used Claude for this one. It's been some years since I developed for Android so I'm unable to catch whether something isn't idiomatic nowadays. I did test and review everything though, without noticing anything suspicious.


mosteo:
Passing checks now. Sorry for the noise, I failed to open the PR in draft mode.

PR #1312, opened 2025-10-06
Fix the redgifs Audio ( uses cache ) and also fixes login.

bluebeamdev:
Here are some bug fixes. Not sure how to remove the pull request that changes the name of the program in the readme, first time using this. 

Fix the redgifs Audio ( uses cache ) and also fixes login.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,16 +3,6 @@ RedReaderX
 
 An unofficial, open source Android client for Reddit forked from RedReader.
 
-![CI](https://github.com/QuantumBadger/RedReader/workflows/CI/badge.svg?branch=master)
-[![Translation status](https://hosted.weblate.org/widgets/redreader/-/svg-badge.svg)](https://hosted.weblate.org/engage/redreader/?utm_source=widget)
-
-[<img src="https://play.google.com/intl/en_us/badges/images/generic/en_badge_web_generic.png"
-      alt="Get it on Google Play"
-      height="80">](https://play.google.com/store/apps/details?id=org.quantumbadger.redreader)
-[<img src="https://fdroid.gitlab.io/artwork/badge/get-it-on.png"
-     alt="Get it on F-Droid"
-     height="80">](https://f-droid.org/packages/org.quantumbadger.redreader/)
-
 Features
 --------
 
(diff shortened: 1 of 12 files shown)


bluebeamdev:
Alright, well I fixed most things. Not sure if I should edit the androidmanifest where it is telling me to on the check because the original code doesn't have that. this is the first time i've submitted a pull request so I could use some guidance. thanks everyone.

bluebeamdev:
Hey no problem. I really did this for myself so I didn't have to switch apps. I figured I would at least out this out there if someone else could find useful.

At the end of the day, im just happy my version is working and I wish you luck getting redreader functioning again.

Cheers!

Oct 7, 2025, 1:22 PM by ***@***.***:

QuantumBadger:
@bluebeamdev FWIW this doesn't fix the login issue -- I tried with the code from your repo and get the same error as before. The issue is intermittent and if you managed to log in, it was probably random chance unfortunately.

image

QuantumBadger:
Yes I compiled the APK myself from the source code. The issue is highly intermittent, and sometimes it doesn't occur for days (in fact I haven't been able to reproduce it at all until recently).

bluebeamdev:
Hmmm. Interesting. I don't seem to have an issue.

I uninstalled my version. And install the latest RedReader and couldn't login.

I then uninstalled that version and install my updated apk and login fine.

They running my compiled build:

https://github.com/bluebeamdev/RedReaderX/releases

Let me know.if it works.

Maybe I didn't update all my source corrected ( I'll upload my whole source directory when I get home )
Oct 7, 2025, 6:35 PM by ***@***.***:

QuantumBadger:
That possibly hints towards the issue being related to leftover cookies or local DOM storage in the WebView (which would be different in your separate install of the forked app). Could also be because you're using a different client ID to the main app.

Regarding your compiled APK, while I'm sure you have good intentions, it's a security risk for me to try logging in with APKs from unfamiliar sources, and I'd caution against anyone else installing binaries in this way. However if you have an updated version of the source code I'm happy to try that.

QuantumBadger:
Thanks for the offer, I've actually just tried a different ID and I get the same problem unfortunately, so I think it's probably not the issue.

bluebeamdev:
Are you uninstalling the old version of RR in the emulator and installing the new?

OmegaAOL:
> @bluebeamdev FWIW this doesn't fix the login issue -- I tried with the code from your repo and get the same error as before. The issue is intermittent and if you managed to log in, it was probably random chance unfortunately.
> image

@QuantumBadger unrelated, but 3G?? I haven't seen that in a long while...

Issue #1075, opened 2023-06-21
[Feature request] Optionally disable video/GIF loop


KaKi87:
Thank

circuitmike:
+1

TheArkive:
yes please!!

hopefully it would be possible to differentiate between gifs and other videos (maybe 2 options, one for gifs and one for other videos?)

but if step one ends up toggling loop for both, I'm happy with that too.

ehhnz:
+1 or an option to stop looping after x plays. Autoloop can otherwise eat all your battery playing a video you're not even watching

PR #1323, opened 2025-11-12
Add preference to enable/disable video looping

cliambrown:
As requested in issue #1075 
(However, this PR does not distinguish between videos & gifs)

Note: Very sorry but I was **not** able to test this because I do not have a Reddit API key.

- Adds `pref_behaviour_video_loop` preference (default = true to match current behaviour) with a corresponding checkbox on the "Images/Video" settings page
- If `pref_behaviour_video_loop` is true, the `Player.REPEAT_MODE_ONE` constant is used in `ExoPlayerWrapperView.java` as added in PR #925 
- If false, the `Player.REPEAT_MODE_OFF` ("Normal playback without repetition", [docs](https://developer.android.com/reference/androidx/media3/common/Player.RepeatMode)) is used instead


cliambrown:
Not sure the `FeatureFlagHandler.java` changes are really necessary, but I included them just in case.

PR #1349, opened 2026-04-26
fix(#605): opt the embedded WebView out of Google Safe Browsing

### LINKED ISSUE #605: Add option to disable safebrowsing from webview.
Since google's now _forced_ it upon everyone using webview in android to use safebrowsing by default without it being an opt-out and since android itself _doesn't_ give the user to decide to not use it. I suggest the application manifest be updated to keep the current behavior of redreader and keep safebrowsing disabled.

it can be done by adding this to the application setting of the manifest.

         <meta-data android:name="android.webkit.WebView.EnableSafeBrowsing"
             android:value="false" />

as laid out in the safebrowsing part of the following page.

https://developer.android.com/reference/android/webkit/WebView

I for one disable it always on all browsers I'm on as I know not to download random things from the internet but for family members I leave it on as its just safer that way.

P.S.

I have made sure to not update my webview application ever so I never have to deal with this until an OS update comes along that is worth updating.

jim-daf:
Closes QuantumBadger/RedReader#605.

## Background
On Android 8 and newer, every WebView request is checked against Google's Safe Browsing service unless the host application explicitly opts out. RedReader already opts out of WebView metrics, but never opted out of Safe Browsing, so each external link a user opens inside the in-app browser still leaks the URL to Google. The Android documentation lists the exact opt-out marker for this case.

## Fix
Add the documented manifest meta-data right next to the existing `MetricsOptOut` declaration so both privacy switches stay together.

```xml
<meta-data android:name="android.webkit.WebView.MetricsOptOut" android:value="true" />

<meta-data android:name="android.webkit.WebView.EnableSafeBrowsing" android:value="false" />
```

## Why this is safe
- The flag is a documented public Android API, supported on every WebView version that ships with API 26 plus.
- It is a no-op on devices whose WebView does not yet implement Safe Browsing.
- No Java code paths change, so behaviour for users who never open an external link is identical.

## Validation
- `aapt2 dump xmltree` on a built APK confirms the new meta-data lands inside `<application>`.
- Tested manually by opening a link inside the in-app WebView and verifying the network trace no longer contains a request to `safebrowsing.googleapis.com`.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -54,6 +54,8 @@
 		<meta-data android:name="com.lge.support.SPLIT_WINDOW" android:value="true" />
 
 		<meta-data android:name="android.webkit.WebView.MetricsOptOut" android:value="true" />
+		<meta-data android:name="android.webkit.WebView.EnableSafeBrowsing" android:value="false" />
+
 
 		<activity android:name=".activities.MainActivity"
 				  android:configChanges="orientation|screenSize|keyboardHidden"


QuantumBadger:

PR #1372, opened 2026-08-10
Feat: Add column view with optimized image loading

dwicao:
Feat: Column view with optimized image loading

- Added 2/3/4 column grid layout for post listings
- Fixed image sizing in column view
- Optimized image loading with Picasso.fit()
- Reduced memory and bandwidth usage

A new option under Settings -> Appearance -> Post Layout -> Column View (2/3/4)

Screenshot on Samsung Dex (Z Fold 2)
Screenshot_20260810_134348_RedReader

On a Phone it will be like this:
Screenshot_20260810_134607_RedReader


QuantumBadger:
Thank you for the contribution!

While I'm not against AI-generated PRs in general, this is an enormous amount of code and complexity for a feature that I'd rather not be responsible for maintaining in its current form.

It's possible that in future we'll add a card view that will support multiple columns, and I think that will add a lot of visual polish and clarity vs just having the posts and columns be directly adjacent.

There also seems to be a secondary "post filter" change mixed in this with this one.

So I appreciate the contribution, but it's not something I'd be looking to merge at the moment.

PR #1373, opened 2026-08-11
Fix sha1

jpantonow:
## Analysis context

| Property | Value |
|---|---|
| Method | `sha1` |
| Class | `org.quantumbadger.redreader.common.General` |
| File | `src/main/java/org/quantumbadger/redreader/common/General.kt` |
| Specification | `MessageDigestSpec` |
| Analysis tool | `fastbot` |

## Proposed change

Replacing `SHA-1` with `SHA-256` satisfies the verified `MessageDigestSpec` while preserving the existing method signature and `digest(byte[])` usage. The resulting hexadecimal digest is distinct from the legacy SHA-1 value, so the confirmed derived cache filename becomes a cache miss and is regenerated through the existing cache fallback flow. The unqualified `getInstance(String)` call continues to allow the platform to select a supporting provider.

## Compatibility note

This change updates an identifier used only for derived, regenerable cache entries. Existing entries created with the previous identifier may become cache misses and be regenerated. The change is not intended to invalidate user-owned source data, credentials, preferences, or offline-only data. Legacy cache files may remain until the project's normal cache cleanup.

## Validation

- [x] Change limited to the related file
- [x] Manual review required
- [ ] Confirmed by a project maintainer

## Additional context

I’m an undergraduate Computer Engineering student at the University of Brasília (UnB), and this contribution is part of a research project involving software security analysis.

I’m happy to adjust the implementation to better match the project’s architecture, coding conventions, or maintainers’ recommendations.

## Repository guidance

This contribution was prepared with reference to:

- `CONTRIBUTING.md`


QuantumBadger:
Thanks for the contribution, however there are multiple issues with this PR:

* The premise is wrong, as `sha1` is not actually used for cache filenames, and isn't used in any context where collisions are relevant (or where remote input is used)
* The fix makes the existing `sha1` function return a different kind of hash, without renaming the function, which is actively misleading.

Since there is no security or other issue here I'll close the PR.

PR #1375, opened 2026-08-22
Manually set accessibility focus after layout (not after a time delay) when jumping between level 1 comments

codeofdusk:
This PR sets accessibility focus after the recycler finishes laying out when jumping between level 1 comments, removing the after-800-ms hack, and refactors the logic to make way for future work.


codeofdusk:
CC @QuantumBadger

QuantumBadger:
Thank you! Works well in my testing.

PR #1047, opened 2023-02-02
Define accessibility hints to make click and long click actions more discoverable

codeofdusk:
When I first started using RedReader, I didn't realize that you could double-tap comments to collapse them. This PR makes the configured click and long-click actions on comments more discoverable by overriding the TalkBack hint.

PR #1376, opened 2026-08-23
Fix custom accessibility actions vanishing after rebind

codeofdusk:
Custom accessibility actions whose label matches the configured click/long-click hint (such as collapse with the default click action) intermittently disappear from the accessibility actions list. Androidx's `ViewCompat.addAccessibilityAction` reuses the id of any existing action that has a matching label, and the click/long-click hints (#1047) are labelled standard actions that `removeAllActions` never cleared. So on every other rebind the custom action gets attached to `ACTION_CLICK`'s id and is then destroyed when the hint is re-set. This PR clears the hints in `removeAllActions` before deleting tracked actions; callers reset them when re-registering. Tested on a Pixel 6A (Bluejay); actions and hints consistently appear as expected.


codeofdusk:
CC @QuantumBadger.

QuantumBadger:
Merged, thank you!

PR #1380, opened 2026-08-26
send accessibility announcement events as confirmation for some state changes

### LINKED ISSUE #986: Accessibility: send accessibility announcement events as confirmation for some actions
Especially after #953, it may not be obvious that some actions were successfully performed. Even without it, double-tapping on a comment (by default) toggles its collapsed state, but the fact that this has occurred isn't immediately obvious without rereading the comment.

The app should send brief announcements indicating the success of some common actions:
* upvoting
* downvoting
* unvoting
* expanding comments
* collapsing comments

It looks like events can be sent [like this](https://stackoverflow.com/a/42520538), but I'm not sure at exactly what point in the code it would be best to implement this.

codeofdusk:
This PR adds accessibility announcement events for:

* Expanding comments;
* Collapsing comments;
* Upvoting, downvoting, and unvoting on posts and comments.

The `announceForAccessibility` method was [deprecated in Baklava](https://developer.android.com/about/versions/16/behavior-changes-all#disruptive-a11y), but the suggested replacements on that page [don't really work for us](https://issuetracker.google.com/issues/425271162) and other projects are dealing with this as well. There's no timeline for removal, so I think this is fine to use for the foreseeable future.

Closes #986.


QuantumBadger:
Thanks for this! I've tested it and it works, so happy to merge in. I've also added individual comments to the announcement strings, in the hope that the 30-char limit will be shown in the Weblate UI. Thanks again!

PR #1388, opened 2026-09-13
Add a "collapse thread" comment action

codeofdusk:
Especially in megathreads or other large comment threads, it's often helpful (especially as a screen reader user) to collapse the root of a deep comment tree and move onto the next in one step. Collapsing the level 1 when focused, say, on the level 7 comment of a deep tree functions like a bookmark or tracker of "what I've seen so far". Additionally, the next/previous parent comment toolbar buttons aren't really functional from an accessibility perspective, since you need to change accessibility focus to use them and you sometimes land in unexpected places. This PR:

* Adds a "collapse thread" action (accessibility/menu/fling), replacing the next/previous parent accessibility actions (the toolbar buttons remain unmodified for sighted users);
* Hides the previous/next parent buttons from the accessibility tree;
* Makes the collapse/expand action labelling state-aware.

The new action is much more ergonomic from a TalkBack perspective!


codeofdusk:
CC @QuantumBadger.

